PR #14, opened 2019-07-11
Adding GA and Angulartics

pstjean:


src/assets/google-analytics.js
```diff
@@ -0,0 +1,16 @@
...
+    m = s.getElementsByTagName(o)[0];
+  a.async = 1;
+  a.src = g;
+  m.parentNode.insertBefore(a, m);
+}
+
+if (window.ENV && window.ENV.GA_KEY) {
+  initGoogleAnalytics(window, document, 'script', 'https://www.google-analytics.com/analytics.js', 'ga');
+  ga('create', window.ENV.GA_KEY, 'auto');
+}
```
sandikbarr, line 16: Does this need to get added to `index.pug`? Looks like we're including it there in Publish and Metrics https://github.com/PRX/publish.prx.org/blob/master/src/index.pug#L21

sandikbarr (changes requested):
think you need to include that `google-analytics.js` files in `index.pug`

src/assets/google-analytics.js
pstjean: Yep, totally. I was thinking the PRX node server iterated through assets and did this automatically for some reason.

sandikbarr (approved):
👍

Issue #21, opened 2019-08-05
Campaign Dashboard Filtering


sandikbarr:
* Filter the campaign dashboard by
  * [x] Flight Date
  * [x] Flight Show
  * [x] Flight Zones
  * [ ] Flight Priority
  * [x] Flight Geolocation 
  * [ ] Campaign Name
  * [x] Advertiser
  * [x] Campaign Rep

* [x] Paging not yet working in API when implemented, also complete/enhance paging.

* [x] Filters should apply to route and when navigating back to the home page in the application or using the back button, the filters should still be applied

sandikbarr:
Hey @cavis, thought you grabbed this ticket for the API, so I created another one #28. Do we have a duplicate or did you mean to close this with the augury filter facets?

cavis:
Whoops!  Didn't mean to close this one ... I think I connected that PR to this Grove issue instead of an Augury one.

cavis:
Also, paging should all work with these filter params.  And they can all be found here: https://github.com/PRX/augury.prx.org/blob/master/app/controllers/api/campaigns_controller.rb#L5

sandikbarr:
K, I'm gonna take this one back and close #28 as dupe. Yeah, I think it was just that paging wasn't implemented yet when I added it on the front end.

PR #23, opened 2019-08-06
campaign form

sandikbarr:


.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
```
pstjean, line 1: Prettier! 🎉

src/app/campaign/form/campaign-form.component.ts
```diff
@@ -0,0 +1,92 @@
...
+      </prx-fancy-field>
+
+    </form>
+  `
+})
+export class CampaignFormComponent implements OnInit {
+  advertiserOptions$: Observable<(string | number)[][]>;
+  accountOptions$: Observable<any[][]>;
+  campaign: CampaignModel;
+  typeOptions = [
```
pstjean: I tend to `Object.freeze` this type of thing

src/app/campaign/list/campaign-list-paging.component.ts
```diff
@@ -0,0 +1,44 @@
...
+        [disabled]="isPageActive(1)"
+        (click)="!isPageActive(1) && showPage.emit(1)"
+        title="Page 1">|«</button>
+      <button class="pager"
+        [disabled]="isPageActive(1)"
+        (click)="!isPageActive(1) && showPage.emit(currentPage - 1)"
+        title="Page {{ currentPage > 1 ? currentPage - 1 : 1 }}">«</button>
+
+      <button
+        *ngFor="let x of ' '.repeat(totalPages).split(''), let i = index"
```
pstjean, line 17: This is wild! Took me a second to figure out what was going on. I _think_ `Array.from({length: totalPages}, (_,idx) => idx)` would be clearer here, but maybe that's just because I've never used `String.prototype.repeat` before.

pstjean: Also, no idea if my thing is even valid in Angular templates.

src/app/campaign/service/advertiser.service.ts
```diff
@@ -0,0 +1,69 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject, Observable, of } from 'rxjs';
+import { map, withLatestFrom, share, mergeMap } from 'rxjs/operators';
+import { AuguryService } from '../../core/augury.service';
+import { AdvertiserModel } from '../../shared/model/advertiser.model';
+
+@Injectable()
```
pstjean, line 7: Based on what this services does, I'm guessing it would be ok to make this a singleton and use `{providedIn: 'root'}` here?

src/app/login/login.component.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class LoginComponent {
+
+  errorMsg: string;
+
+  constructor(private router: Router) {}
+
+  loginSuccess() {
+    this.errorMsg = null;
+    this.router.navigate(['/']);
```
pstjean, line 25: 💯

pstjean (changes requested):

src/app/campaign/list/campaign-list-paging.component.ts
sandikbarr: Array.from is not valid in the template. Not having a simple index without an iterator in templates can be annoying sometimes. I think this code will clean itself up once we finish implementing paging with #21 because it won't be displaying all the pages.

src/app/campaign/service/advertiser.service.ts
sandikbarr: Advertisers are on the Campaigns, so I put it in the CampaignModule, which yes, the service would only be included once from the CampaignModule in the app, but as a logical feature separation that's where I felt they belonged. Instead of listing in providers in the CampaignModule, I could list the services as `{providedIn: CampaignModule}'.  Both ways are supported, and I'd rather stick with what we have in other applications with the providers array on the feature modules for now.

Maybe the UserService would be a good candidate for `{providedIn: 'root'}` since that one is in CodeModule.

I kinda feel like we only made this change in Publish where we had to instead of making it a new standard. If we keep things more samer, everyone will have an easier time knowing what's going on.

src/app/campaign/form/campaign-form.component.ts
sandikbarr: I'm cool with not enforcing object mutation on these select options by runtime crash, but I added `readonly` because that seemed like a more TypeScripty compromise.

pstjean:
Looks great!

PR #39, opened 2019-09-05
Campaign form

pstjean:
This is missing any attempt at final styling (because we want to evaluate Angular Material) and stuffs a bunch of data fetching in this component that will need to be refactored out (because we want to prioritize doing one thing at a time, i.e. getting a reactive form that works in the right place).

### DIFF
--- a/src/app/campaign/campaign-form.component.html
+++ b/src/app/campaign/campaign-form.component.html
@@ -1,0 +1,47 @@
+<form [formGroup]="campaignForm" (ngSubmit)="campaignFormSubmit()">
+  <label>
+      Owner:
+      <select formControlName="accountId">
+        <option *ngFor="let acct of accountOptions$ | async" [ngValue]="acct.value">
+          {{acct.name}}
+        </option>
+      </select>
+    </label>
+
+  <label>
+    Name:
+    <input type="text" formControlName="name" required>
+  </label>
+
+  <label>
+    Type:
+    <select formControlName="type">
+      <option *ngFor="let type of typeOptions" [ngValue]="type.value">
+        {{type.name}}
+      </option>
+    </select>
+  </label>
+
+  <label>
+    Status:
+    <select formControlName="status">
+      <option *ngFor="let status of statusOptions" [ngValue]="status.value">
+        {{status.name}}
+      </option>
+    </select>
+  </label>
+
+  <label>
+    Rep name:
+    <input type="text" formControlName="repName">
+  </label>
+
+  <label>
+    Notes:
+    <textarea formControlName="notes"></textarea>
+  </label>
+
+
+
+  <button type="submit" [disabled]="!campaignForm.valid">Submit</button>
+</form>
--- a/src/app/campaign/campaign-form.component.html
+++ b/src/app/campaign/campaign-form.component.html
@@ -1,21 +1,21 @@
-<form [formGroup]="campaignForm" (ngSubmit)="campaignFormSubmit()">
+<form [formGroup]="campaignForm" (ngSubmit)="campaignFormSubmit()" ngNativeValidate>
   <label>
-      Owner:
-      <select formControlName="accountId">
-        <option *ngFor="let acct of accountOptions$ | async" [ngValue]="acct.value">
-          {{acct.name}}
-        </option>
-      </select>
-    </label>
+    Owner:
+    <select formControlName="accountId" required>
+      <option *ngFor="let acct of accountOptions$ | async" [ngValue]="acct.value">
+        {{acct.name}}
+      </option>
+    </select>
+  </label>
 
   <label>
     Name:
-    <input type="text" formControlName="name" required>
+    <input type="text" formControlName="name" [class.invalid]="name.invalid" required>
   </label>
 
   <label>
     Type:
-    <select formControlName="type">
+    <select formControlName="type" required>
       <option *ngFor="let type of typeOptions" [ngValue]="type.value">
         {{type.name}}
       </option>
@@ -24,7 +24,7 @@
 
   <label>
     Status:
-    <select formControlName="status">
+    <select formControlName="status" required>
       <option *ngFor="let status of statusOptions" [ngValue]="status.value">
         {{status.name}}
       </option>
@@ -33,7 +33,7 @@
 
   <label>
     Rep name:
-    <input type="text" formControlName="repName">
+    <input type="text" formControlName="repName" required>
   </label>
 
   <label>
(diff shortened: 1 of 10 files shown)


src/app/campaign/campaign-form.component.html
```diff
@@ -0,0 +1,54 @@
+<form [formGroup]="campaignForm" (ngSubmit)="campaignFormSubmit()" ngNativeValidate>
+  <label>
+    Owner:
+    <select formControlName="accountId" required>
```
cavis: This may be a more recent change to Augury ... but you'll need to do `set_account_uri` here.  (I disabled setting account-ids directly, in favor of the more HAL-ish approach).

src/app/campaign/campaign-form.component.ts
```diff
@@ -0,0 +1,88 @@
...
+  campaignForm = this.fb.group({
+    accountId: ['', Validators.required],
+    name: ['', Validators.required],
+    type: ['', Validators.required],
+    status: ['', Validators.required],
+    repName: ['', Validators.required],
+    notes: [''],
+    set_advertiser_uri: ['', Validators.required]
+  });
+  get accountId() { return this.campaignForm.get('accountId'); }
```
cavis: Also here `set_account_uri`

src/app/campaign/campaign-form.component.ts
```diff
@@ -0,0 +1,88 @@
...
+              private userService: UserService,
+              private auguryService: AuguryService) { }
+
+  ngOnInit() {
+    this.advertiserOptions$ = this.fetchAdvertisers();
+
+    this.accountOptions$ = this.userService.accounts.pipe(
+      withLatestFrom(this.userService.defaultAccount),
+      map(([accounts, defaultAccount]) => {
+        const defaultAccountOption = {name: defaultAccount['name'], value: defaultAccount.id};
```
cavis: You can get the HAL uri by expanding the link to the document itself:

```
defaultAccount.expand('self')
```

src/app/campaign/campaign-form.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    );
+
+  }
+
+  fetchAdvertisers() {
+    const resourceName = 'advertisers';
+    return this.auguryService.followItems(`prx:${resourceName}`).pipe(
+      map((advertisers) => {
+        return advertisers.map(adv => ({
+          name: adv['name'], value: [this.auguryService.path, resourceName, adv['id']].join('/')
```
cavis: Yeah, here I'd use that `adv.expand('self')` helper too.

src/app/campaign/campaign-form.component.ts
```diff
@@ -0,0 +1,88 @@
...
+        }));
+      })
+    );
+  }
+
+  campaignFormSubmit() {
+    this.auguryService.root.subscribe(rootDoc => {
+      console.log(rootDoc);
+      const resp = rootDoc.create('prx:campaign', {}, this.campaignForm.value);
+      resp.subscribe(res => console.log(res));
```
cavis: Probably a redirect here, eventually?

src/app/campaign/campaign-form.component.ts
```diff
@@ -0,0 +1,88 @@
...
+  get status() { return this.campaignForm.get('status'); }
+  get repName() { return this.campaignForm.get('repName'); }
+  get notes() { return this.campaignForm.get('notes'); }
+  get set_advertiser_uri() { return this.campaignForm.get('set_advertiser_uri'); }
+
+  constructor(private fb: FormBuilder,
+              private userService: UserService,
+              private auguryService: AuguryService) { }
+
+  ngOnInit() {
```
cavis, line 92: Where do values get initialized from the current/existing Campaign?  (Or maybe that's a TODO).

src/app/campaign/campaign-form.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    {name: 'Hold', value: 'hold'},
+    {name: 'Sold', value: 'sold'},
+    {name: 'Approved', value: 'approved'},
+    {name: 'Paused', value: 'paused'},
+    {name: 'Canceled', value: 'canceled'}
+  ];
+
+  campaignForm = this.fb.group({
+    accountId: ['', Validators.required],
+    name: ['', Validators.required],
```
cavis, line 53: Liking this syntax.

cavis (changes requested):
Nice - liking the more functional approach of the FormComponent.

Couple questions, which are probably already TODOs.  Only changes requested are really (1) augury doesn't let you directly set an account id anymore - you have to use the set uri thing.  And (2) the existence of the `.expand('self')` method on HalDoc, everyone's favorite undocumented library in styleguide.

src/app/campaign/campaign-form.component.ts
pstjean: Yeah, that's part of what needed to come over from even more prototype-y work. Have added that after a bit of refactoring.

src/app/campaign/campaign-form.component.ts
```diff
@@ -0,0 +1,161 @@
...
+          resp = cmp.update(this.campaignForm.value);
+        } else {
+          resp = root.create('prx:campaign', {}, this.campaignForm.value);
+        }
+        resp.subscribe(res => {
+          this.toastr.success('Campaign saved');
+          if (!cmp) { this.router.navigate(['/campaign', res.id]); }
+        });
+      })
+    ).subscribe();
```
pstjean: Some nested Observable stuff that I don't like at all, but works for now.

src/app/core/core.env.ts
```diff
@@ -47,6 +47,7 @@ const getNumber = (name: string): number => {
 };
 
 export class Env {
+  public static get CMS_HOST(): string { return getVar('CMS_HOST'); }
```
cavis: May be extraneous now.

src/app/campaign/campaign-form.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(CampaignFormComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  xit('should create', () => {
```
cavis: Would be nice to test the non-builtin functionality (some of the HalMock stuff, to test that the form creates/updates with the correct keys/values).  But can take that as a TODO, or we can work on that in a bit.

cavis (approved):
:+100: one extraneous thing, and some lack of tests.  But otherwise yeah, looks good.

PR #45, opened 2019-09-16
filter by facets

sandikbarr:


src/app/campaign-list/campaign-list.service.ts
```diff
@@ -37,34 +55,37 @@ export interface Campaign {
...
-  };
+  params: CampaignParams = {page: 1, per: 12};
+  facets: Facets;
   total: number;
   count: number;
   error: Error;
-  loading = false;
-  flightsLoaded: number;
   // tslint:disable-next-line: variable-name
   private _campaigns = new BehaviorSubject<{[id: number]: Campaign}>({});
```
sandikbarr, line 69: this._campaigns holds all campaigns that have been requested, so if we did decide to do something with using this application state as a cache, it's here

src/app/campaign-list/campaign-list.service.ts
```diff
@@ -37,34 +55,37 @@ export interface Campaign {
...
+  }
+
   get campaigns(): Observable<Campaign[]> {
     // this._campaigns is a subject of campaign entities (an object with campaignIds mapped to campaign,) convert to array for display
     return this._campaigns.asObservable().pipe(
       map((campaignEntities: {}) => {
-        if (Object.keys(campaignEntities).length) {
-          return Object.keys(campaignEntities)
-            // map to array
+        if (this._currentCampaignIds && this._currentCampaignIds.length) {
```
sandikbarr, line 85: this._currentCampaignIds is an array of campaign Ids that were last requested. Since this._campaigns holds all campaigns that have been requested (during the session), I'm limiting what is returned here for display to the current results.

src/app/campaign-list/list/filter-date.component.scss
```diff
@@ -0,0 +1,10 @@
+/* have to override the global button styles */
```
sandikbarr, line 1: We should really remove global button styles and move them to scoped component styles if we're getting serious about using material components.

src/app/campaign-list/campaign-list.service.ts
cavis: +1

src/app/campaign-list/campaign-list.service.ts
```diff
@@ -76,31 +97,38 @@ export class CampaignListService {
...
-      concatMap((campaignDocs: HalDoc[]) => {
-        if (campaignDocs.length) {
-          this.count = campaignDocs[0]['_count'];
-          this.total = campaignDocs[0]['_total'];
-          this.facets = campaignDocs[0]['_facets'];
-        }
+    this.augury.follow('prx:campaigns', {
+      page,
+      per,
+      ...(filters && {filters})
```
cavis, line 113: /me adds that to my object-spread-syntax toolkit

src/app/campaign-list/campaign-list.service.ts
```diff
@@ -76,31 +97,38 @@ export class CampaignListService {
...
+    }).pipe(
+      switchMap(result => {
+        this.count = +result['count'];
+        this.total = +result['total'];
+        this.facets = result['facets'];
+        return result.followList('prx:items');
+      }),
+      switchMap((campaignDocs: HalDoc[]) => {
+        this._campaigns.next(campaignDocs.reduce((acc, doc) => {
```
cavis, line 122: One neat trick i picked up from Peter ... 

```js
this._campaigns.next(campaignDocs.reduce((acc, doc) => ({...acc, [doc.id]: {loading: true}}), {}));
```

(That may just make me look desperate to write 1-liners).

src/app/campaign-list/campaign-list.service.ts
```diff
@@ -146,4 +173,66 @@ export class CampaignListService {
     },
     err => this.error = err);
   }
+
+  getFilters(params: CampaignParams) {
```
cavis, line 172: This is nice ... wrapping up the specifics of filters.

src/app/campaign-list/card/campaign-flight-targets.pipe.ts
```diff
@@ -7,10 +7,10 @@ import { Flight } from '../campaign-list.service';
 export class CampaignFlightTargetsPipe implements PipeTransform {
   transform(flights: Flight[]): string {
     if (flights && flights.some(flight => flight.targets && flight.targets.length > 0)) {
-      return flights.reduce((acc, flight) => {
-        return acc += flight.targets && (acc && ', ') +
-          flight.targets.map(t => t.label).join(', ');
-      }, '');
+      const targets = flights.reduce((fAcc, flight) => {
+        return {...fAcc, ...flight.targets.reduce((tAcc, target) => ({...tAcc, [target.label]: target.label}), {})};
```
cavis, line 11: Oof ... double reducers!  I guess there's no good way around it ... just wish typescript had a better utility for doing this.

src/app/campaign-list/list/campaign-list.component.ts
```diff
@@ -0,0 +1,138 @@
...
+  template: `
+    <grove-campaign-filter
+      [params]="params" [facets]="facets" (campaignListParams)="routeToParams($event)">
+    </grove-campaign-filter>
+    <h2>Campaigns</h2>
+    <ul>
+      <li *ngFor="let campaign of campaigns$ | async">
+        <grove-campaign-card [campaign]="campaign"></grove-campaign-card>
+      </li>
+      <ng-container *ngIf="loading$ | async as loading">
```
cavis, line 17: Slick way of doing loading indicators!

src/app/campaign-list/list/campaign-list.component.ts
```diff
@@ -0,0 +1,138 @@
...
+  get facets(): Facets {
+    return this.campaignListService.facets;
+  }
+
+  routeToParams(partialParams: CampaignParams) {
+    let { page, advertiser, podcast, status, type, text, representative } = partialParams;
+
+    // this function takes partial parameters (what changed)
+    // mix in the existing this.params unless property was explicitly set in partialParams
+    if (!partialParams.hasOwnProperty('page') && this.params.page) {
```
cavis, line 78: Would this be any different than just checking `page === undefined` or something?

src/app/campaign-list/list/filter-date.component.scss
cavis: I also saw that madness, in the datepicker.  Good monkey patch!

src/app/campaign-list/list/filter-facet.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class FilterFacetComponent {
+  @Input() facetName: string;
+  @Input() multiple: boolean;
+  @Input() options: Facet[];
+  @Input()
+  set selectedOptions(values: number | string | number[] | string[]) {
+    this._selectedOptions = values;
+    this.selectedOptionsChange.emit(values);
```
cavis, line 27: This is a cool way to emit whenever the field changes... may steal this...

src/app/campaign-list/campaign-list.service.ts
cavis: Oh, and should definitely throw a `zoom: 'prx:flights,prx:advertiser'` in here (it's working now).  Cuts down on the remote requests quite a bit.

cavis (approved):
👍 looks good to me.  One `?zoom` to be added to a request, but that's about it.

I did notice while testing ... a bunch of our text fields (including the date fields) do weird autocomplete for me.  Maybe we should just disable that throughout most of the app?  I get cases where it's trying to fill non-date-values into the date filters and such.

src/app/campaign-list/card/campaign-flight-targets.pipe.ts
sandikbarr: yeah, a little ugly. we could turn it into a stream and use distinct on it, stream all the things! yeah, maybe this is fine, I'm getting carried away

src/app/campaign-list/list/campaign-list.component.ts
sandikbarr: They arent really showing now that I added the zoom, but it did seem cool at the time! I was trying to do some kind of skeleton interface while loading, but maybe we'll decide something else for loading indication once the design settles in.

src/app/campaign-list/list/campaign-list.component.ts
sandikbarr: I didn't love this, but the campaign-filter.component emits `{type: undefined, page: 1}` when the None option is selected from a select (because it doesn't have a value, which is what material recommends). Maybe not great naming, but partialParams won't contain the full route parameters, just what changed in the filters. So I needed a way to determine if the partialParams was meant to clear the value (hasOwnProperty) or if it should get the value from the full route `this.params`. Sometimes the negative logic also makes things less readable for me, so I think that's partly why this is confusing.

src/app/campaign-list/list/filter-facet.component.ts
sandikbarr: yeah, seemed like I had to use this here because the material select uses that double binding on the value

sandikbarr:
pushed a couple changes for zoom, using the one line reduce, and removed a bad sort. planning to merge on Monday

Issue #11, opened 2019-07-10
Flight Form Component


sandikbarr:
Basic flight fields, as seen [somewhere in Sketch](https://sketch.cloud/s/On4Gq/a/ReEvRx/play).  Note that this doesn't support saving the flight yet.  Just some form fields with remote values.

- [x] "+ Add a Flight" link/button in left column, under the campaign.  Display the flight name as a tab.
- [ ] ~Some TBD way to remove the flight from the campaign~ deal with it!
- [x] "Name" field (required text w/length validation)
- [x] "Series" dropdown (required, single-select)
  - Pull from augury `prx:inventories`, with a `zoom: 0` and somehow defeat paging with a `per: 999` or something.
  - On select, refresh the list of "Zones" from the inventory resource.  TBD where this is in the API - PRX/augury.prx.org#91
- [x] "Start Date" (require, date)
- [x] "End Date" (required, date, validate > start)
- [x] "Zone" dropdown (required, single-select)
  - Disabled until the Series is selected
  - Clear when the Series changes

PR #46, opened 2019-09-17
Campaign flight layout

pstjean:
This is for (most-of-but-also-more-than) #11.  Still need to add the "Zone" dropdown to the flight form, but can tackle that separately. And @brandonhundt wants some of the dropdowns changed to autocomplete fields.

Overview:

- Add `CampaignState` and `FlightState` models to represent the current state of the form(s).  And the CampaignState has many FlightStates, so we can capture all of them in the single/global "save" button 
- Add a `CampaignService` to handle the remote loading/setting of campaigns and flights, and translating them to models.
- Add a `CampaignStore` to track the current states and manage the complexity of chain-creating/updating them
- Add form components + containers for the campaign and flights

(This PR is a great candidate for squash-merging)


src/app/campaign/flight/flight.component.html
```diff
@@ -1,26 +1,41 @@
...
       <mat-datepicker-toggle matSuffix [for]="endPicker"></mat-datepicker-toggle>
       <mat-datepicker #endPicker></mat-datepicker>
     </mat-form-field>
-    <mat-form-field>
+    <mat-form-field appearance="outline">
       <mat-label>Series</mat-label>
-      <mat-select formControlName="set_inventory_uri" required>
-        <mat-option *ngFor="let inv of inventory" [value]="inv.self_uri">{{ inv.podcastTitle }}</mat-option>
-      </mat-select>
+      <input
```
brandonhundt: This isn't filtering and its printing the URL to the selected item and not the name of it.

src/app/campaign/form/campaign-form.component.html
```diff
@@ -1,43 +1,49 @@
...
-        <mat-option *ngFor="let acct of accounts" [value]="acct.self_uri">{{ acct.name }}</mat-option>
-      </mat-select>
+    <mat-form-field appearance="outline">
+      <mat-label>Campaign Name</mat-label>
+      <input matInput placeholder="Campaign Name" formControlName="name" required />
     </mat-form-field>
 
-    <mat-form-field class="example-full-width">
-      <input matInput placeholder="Name" formControlName="name" required />
+    <mat-form-field appearance="outline">
```
brandonhundt, line 8: This isn't filtering and its printing the URL to the selected item and not the name of it.

See the filter autocomplete example: https://material.angular.io/components/autocomplete/examples

src/index.pug
```diff
@@ -7,7 +7,7 @@ html(lang='en')
     meta(name='viewport' content='width=device-width, initial-scale=1')
     for style in css
       link(rel='stylesheet' href=style)
-    link(rel='stylesheet' href='https://fonts.googleapis.com/css?family=Roboto:300,400,500')
+    link(rel='stylesheet' href='https://fonts.googleapis.com/css?family=Roboto:300,400,500,700')
```
brandonhundt, line 10: I needed a bold font option!

src/app/campaign/form/campaign-form.component.scss
```diff
@@ -8,3 +8,13 @@ mat-card form > * {
   max-width: 600px;
   width: 100%;
 }
+
+.mat-select-value-text,
+.mat-input-element {
+  font-size: 1rem;
+  color: #fff;
```
brandonhundt: Whoops. change to `#555`

src/app/app.routing.module.ts
```diff
@@ -4,23 +4,21 @@ import { Routes, RouterModule } from '@angular/router';
...
 import { LoginComponent } from './login/login.component';
+import { TokenAuthComponent } from './tokenauth/tokenauth.component';
 
 const routes: Routes = [
   { path: 'permission-denied', component: AuthorizationComponent },
-  { path: '',          component: HomeComponent,  canActivate: [AuthGuard] },
-  { path: 'login',     component: LoginComponent, canActivate: [UnauthGuard] }
+  { path: '', component: HomeComponent, canActivate: [AuthGuard] },
+  { path: 'login', component: LoginComponent, canActivate: [UnauthGuard] },
+  { path: 'tokenauth', component: TokenAuthComponent, canActivate: [UnauthGuard] }
```
sandikbarr, line 15: What is `tokenauth`, is this some testing thing? Should we feature flag it somehow?

src/app/campaign/campaign.component.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    };
+  }
+
+  beforeEach(() => {
+    routeId = new ReplaySubject(1);
+    route = { paramMap: routeId.pipe(map(id => ({ get: jest.fn(() => id) }))) } as any;
+    router = <any>{ navigate: jest.fn(), url: '/campaign/new/flight/9999' };
+    toastrService = <any>{ success: jest.fn() };
+    campaignState = new ReplaySubject(1);
+    campaignStoreService = {
```
sandikbarr, line 48: Is this the only place in the app the CampaignStoreService will ever be injected? If we expect to need it elsewhere, I would suggest doing a separate mock class implementation so we don't repeat ourselves in tests like we have gotten into the habit of doing every time we need to mock something.

src/app/campaign/campaign.component.ts
```diff
@@ -0,0 +1,95 @@
...
+    private router: Router,
+    private toastr: ToastrService,
+    public campaignStoreService: CampaignStoreService
+  ) {
+    this.route.paramMap.subscribe((params: ParamMap) => {
+      this.campaignStoreService.load(params.get('id'));
+    });
+    this.campaignFlights$ = this.campaignStoreService.flights$.pipe(
+      map(flightStates => {
+        return Object.keys(flightStates).map(id => {
```
sandikbarr, line 56: Does the order of these flights matter? JavaScript objects (flight ids mapped to flights) don't preserve order.

src/app/campaign/form/campaign-form.component.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  constructor(private fb: FormBuilder) {}
+
+  ngOnInit() {
+    this.campaignForm.valueChanges.subscribe(cmp => {
+      this.formStatusChanged(cmp);
+    });
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
```
sandikbarr: Not requiring a change here, just offering a best practices (that I know of) idea- Instead of ngOnChanges (the slippery slope of change detection performance) it seems like @Input campaign could have been implemented as a getter/setter pair with the setter calling the this.updateCampaignForm function.

src/app/core/advertiser/advertiser.service.ts
```diff
@@ -0,0 +1,32 @@
...
+  name: string;
+  self_uri: string;
+}
+
+@Injectable()
+export class AdvertiserService {
+  constructor(private augury: AuguryService) {}
+
+  listAdvertisers(params = {}): Observable<Advertiser[]> {
+    return this.augury.followItems('prx:advertisers', params).pipe(map(docs => this.docsToAdvertisers(docs)));
```
sandikbarr, line 18: 👍excellent for now, but I'm replacing this with a stateful service so newly added advertisers can be included in the list

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -0,0 +1,136 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class CampaignStoreService {
+  readonly campaign$ = new ReplaySubject<CampaignState>(1);
+  readonly flights$ = this.campaign$.pipe(map(c => c.flights));
+  readonly remoteCampaign$ = this.campaign$.pipe(map(state => state && state.remoteCampaign));
+  readonly localCampaign$ = this.campaign$.pipe(map(state => state && state.localCampaign));
+
+  get campaign(): Observable<CampaignState> {
+    return this.campaign$.pipe(first());
```
sandikbarr, line 17: They also have a function to return a Subject as an Observable: `this.campaign$.asObservable()`

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -0,0 +1,136 @@
+import { Injectable } from '@angular/core';
+import { CampaignService } from './campaign.service';
+import { CampaignState, FlightState, CampaignStateChanges, Campaign } from './campaign.models';
+import { ReplaySubject, Observable, forkJoin, of } from 'rxjs';
+import { map, first, switchMap } from 'rxjs/operators';
+
+@Injectable({ providedIn: 'root' })
+export class CampaignStoreService {
+  readonly campaign$ = new ReplaySubject<CampaignState>(1);
```
sandikbarr: maybe you meant `private` here instead of `readonly`? You don't typically expose the Subject, in general making them private and exposing them as an Observable prevents consumers from `next()`ing onto it.

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -0,0 +1,136 @@
...
+          set_advertiser_uri: null
+        },
+        flights: {},
+        changed: false,
+        valid: false
+      };
+      this.campaign$.next(newState);
+      return of(newState);
+    } else {
+      const lazyState = new ReplaySubject<CampaignState>(1);
```
sandikbarr: nit: I'm not sure about this temporal lazyState? can consumers not use the campaign state (exposed asObservable)?

sandikbarr (approved):
Some suggestions, but looking good here

src/app/app.routing.module.ts
cavis: Hmm... yeah, looks like it's a workaround for logging in for Cypress tests.  So it would be needed in prod, if we want to run smoketests there or something.

I'm going to say we leave it for now, but hopefully factor it out someday (doing a "real" login via id.prx.org, then opening Grove is probably a better test that things work).

src/app/campaign/campaign.component.ts
cavis: +1.  Will probably need that, but will tackle separately (still needs some UX feedback on how to sort).

src/app/campaign/campaign.component.spec.ts
cavis: Right now, it looks like _most_ of the campaign store service is only used here.  So I agree, but I'm going to push that to the next (hopefully smaller) feature PR that ends up using this store.  And only partially because i'm tired of messing around with tests, and this PR needs to merge!

src/app/campaign/form/campaign-form.component.ts
cavis: Oooh, that's a great pattern that I didn't know about.  Going to do that!

src/app/core/advertiser/advertiser.service.ts
cavis: +1

src/app/core/campaign/campaign-store.service.ts
cavis: Done!

src/app/core/campaign/campaign-store.service.ts
cavis: Updated to be actually-read-only getters.

src/app/core/campaign/campaign-store.service.ts
cavis: Converted to `share()`d observables.

PR #54, opened 2019-09-25
advertiser autocomplete

sandikbarr:
Adds a new autocomplete component using Angular Material autocomplete.
* The advertiser list is populated from the advertiser service, a stateful service that allows new advertisers to be added
* when user types into the input:
  * a filtered list is shown
  * a button that allows the user to add an Advertiser if it does not match one in the list
  * a new advertiser may be submitted by clicking the 'Add Advertiser' button or by pressing the enter key
* new advertisers are emitted up to the campaign container which is connected to the advertiser service
  * the advertiser service posts a new advertiser. that post is multicast so both:
    * the service can listen for it to add the new advertiser to the advertiser list state
    * and the campaign container can update the advertiser on the campaign in the campaign service which then trickles down in to the wired up form


src/app/campaign-list/campaign-list.module.ts
```diff
@@ -40,7 +44,12 @@ import { CampaignListService } from './campaign-list.service';
     CommonModule,
     PagingModule,
     RouterModule,
-    SharedModule
+    SharedModule,
```
sandikbarr, line 47: I had tried to import the material components as exported from SharedModule, but I kept getting errors about a date provider even though MatNativeDateModule was being imported. I moved back to importing material components in component closer to where we are using them. If you like Angular modules as much as I do, don't worry it's probably going to change again.

src/app/campaign/form/campaign-form.component.ts
```diff
@@ -95,17 +96,21 @@ export class CampaignFormComponent implements OnInit {
...
-        name,
-        type,
-        status,
-        repName,
-        notes,
-        set_advertiser_uri
-      },
-      { emitEvent: false }
-    );
+    // when user types an advertiser name match into the advertiser autoselect, set advertiser by value (URI)
```
sandikbarr: This method runs every keydown in the autocomplete, so doing the campaignForm.reset on all fields for every update was causing the advertiser field to clear out. I moved over to this `patchValue` with (very important) options `{emitEvent: false, onlySelf: true}`  and conditionally only included fields that were present.

When typed in, the advertiser value is a "name" but when selected from the autocomplete dropdown, the value is the URI, so this sets the advertiser whether they typed the full matching value (without selecting) or selected one. (There's maybe some further weird UX with autocomplete like it not being easy to opening the options without filtering them, I'm sure we'll talk about it more.)

src/app/home/home.component.spec.ts
```diff
@@ -1,45 +1,49 @@
...
-import { NoopAnimationsModule } from '@angular/platform-browser/animations';
-
-import { SharedModule } from '../shared/shared.module';
 
 import { MockHalService } from 'ngx-prx-styleguide';
-import { CampaignListService } from '../campaign-list/campaign-list.service';
-import { CampaignListModule } from '../campaign-list/campaign-list.module';
+import { CampaignListService, CampaignParams } from '../campaign-list/campaign-list.service';
 import { CampaignListServiceMock, params } from '../campaign-list/campaign-list.service.mock';
 import { HomeComponent } from './home.component';
```
sandikbarr, line 11: Thought I had the material imports sorted out and coming from the SharedModule, but this test is where things really fell apart. (It's just such a pain to reimport everything everywhere and keep track of it all.) So what I did instead was go back to importing material close to where it is used and isolate this parent component test it by mocking out the component it contains.

src/app/shared/autocomplete/autocomplete.component.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  optionName(value?: string): string {
+    if (value && this.options) {
+      const option = this.options.find(opt => opt.value === value);
+      return option && option.name;
+    }
+  }
+
+  // on input: (keydown.enter)="enterPressed($event)"
```
sandikbarr: I haven't decided whether or not I'm including this event trapping. It seems to be an error related to the last pass browser extension when you keydown.enter on the autocomplete.

package.json
```diff
@@ -6,6 +6,7 @@
     "start": "prx-ng-serve dev",
     "build": "ng build --prod --aot",
     "test": "jest --watch",
+    "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
```
sandikbarr, line 9: convenience script for attaching the chrome node inspector

cavis: +100

src/app/campaign-list/campaign-list.module.ts
cavis: I also saw there used to be a `MaterialModule` you could just import all of for testing.  But that would be too easy.

src/app/core/advertiser/advertiser.service.ts
```diff
@@ -1,32 +1,86 @@
...
+    );
+  }
+
+  addAdvertiser(name: string): Observable<{id: number, name: string, set_advertiser_uri: string}> {
+    const post = this.auguryService.root.pipe(
+      mergeMap(root => {
+        return root.create('prx:advertiser', {}, {name});
+      }),
+      // publish + refCount multicasts the Observable (makes it hot/shareable) and gives a complete notification to any late subscribers
+      // unlike share which uses a subject factory so late subscribers cause a retry (duplicating the HTTP POST)
```
cavis, line 66: Worth adding to styleguide/haldoc?

src/app/core/advertiser/advertiser.service.ts
```diff
@@ -1,32 +1,86 @@
...
+          .map(id => advertisers[id])
+          .sort((a, b) => a.name.localeCompare(b.name));
+      })
+    );
   }
 
-  docsToAdvertisers(docs: HalDoc[]): Advertiser[] {
-    return docs.map(this.docToAdvertiser);
+  loadAdvertisers() {
+    this.auguryService.followItems(`prx:advertisers`, {per: 999}).pipe(
```
cavis: Optimization: have the campaign service also `zoom: 'prx:advertiser'` so we can load the field right away, instead of waiting for this call to complete.

src/app/campaign/form/campaign-form.component.ts
```diff
@@ -95,17 +96,21 @@ export class CampaignFormComponent implements OnInit {
...
+    const findAdvertiserByNameOrURI = set_advertiser_uri && this.advertisers &&
```
cavis: Already-known-todo: if you type in an existing advertiser name and save, it tries to PUT the name instead of the uri.

src/app/campaign/form/campaign-form.component.ts
cavis: +1 makes sense.

src/app/shared/autocomplete/autocomplete.component.ts
```diff
@@ -0,0 +1,78 @@
...
+    event.preventDefault();
+    event.stopPropagation();
+    const value = this.formGroup.get(this.controlName).value;
+    if (value && !this.findExistingOtion(value)) {
+      this.addNewOption(value);
+    }
+  }
+
+  findExistingOtion(name: string) {
+    return this.options.find(opt => opt.name === name || opt.value === name);
```
cavis, line 70: Worth noting that in Augury, advertiser names aren't actually unique.  Not a huge deal here, since we probably want them to treat them as unique.

But I do wonder if we'll ever run into that case, where we actually want non-unique values.  I guess we'd need _something_ unique in the display value, if the user is expected to pick between them.

cavis (changes requested):
One bug where typing-but-not-selecting a name PUTs the display value.  But otherwise +202.

I have seen some intermittent issues where dropdowns/autocompletes aren't populating at all on page load.  I've been wondering if it's a race condition, but haven't been able to pin it down yet.  Will keep an eye out for it!

src/app/core/advertiser/advertiser.service.ts
sandikbarr: But that wouldn't give us all the advertisers, just the campaign advertiser, right? (Also campaign already embeds the advertiser by default, but it seems like we only needed the uri on the campaign model that then maps to this list in order to show names on the form.)

cavis: The campaign (+ advertiser) always loads first.  So to avoid the form flashing, I guess I'm saying we could pre-populate this store with just the 1 advertiser.  Then later, add _all_ the advertisers but dedup the one we already added (by id/uri).

sandikbarr: Seems like that could work. I'd kind of like to figure out why the form isn't just displaying the value reactively when it gets advertisers, seems like that should just work in this scenario if we set everything up.

sandikbarr:
Still going to look at the race condition issue with the form not populating, but this latest commit fixes the typed advertiser value at least.

src/app/core/advertiser/advertiser.service.ts
sandikbarr: So I tried this, but I don't really like it, too messy. The campaign (api) service loads the single advertiser with the campaign, which it returns back to the campaign store then the campaign store has to update the advertiser store. If we _have_ to go this route, I guess but I am just not feeling it so far.

src/app/core/advertiser/advertiser.service.ts
sandikbarr: I might be nice as an option to have a shareable Observable, but also http requests aren't shareable by default so maybe that's not the place for it?

src/app/shared/autocomplete/autocomplete.component.ts
sandikbarr: it's definitely not a perfect system right now

src/app/core/advertiser/advertiser.service.ts
sandikbarr: I think what we should do instead of tangling up these stores is load the data sequentially to resolve the race condition. Load the accounts and advertisers, then load the campaign.

sandikbarr: I set it up to load advertisers then campaign. I left accounts alone for now just trying to focus on one thing at a time.

cavis: +1

cavis (approved):
👍 good solution, just loading the autocomplete fields first.  Should be pretty quick.

Issue #32, opened 2019-08-30
Flight Inventory expand-table


cavis:
After #31.

Allow expanding/collapsing weeks into days in the Inventory table, as seen [somewhere in Sketch](https://sketch.cloud/s/On4Gq/a/ReEvRx/play).

- [x] Hovering a week-row shows the expand/collapse icon on the right side of the row
- [x] Clicking icon expands/collapses the daily values

Issue #30, opened 2019-08-30
Flight Inventory Table


cavis:
After #11 and PRX/augury.prx.org#92.

As seen [somewhere in Sketch](https://sketch.cloud/s/On4Gq/a/ReEvRx/play), this is the table on the right side of the page.  Just the "Week" and "Available" columns.

- [x] Display a helpful "fill out the form" message until the required fields are all present
  - (Series, Start Date, End Date, Zone Name)
- [x] When fields change, query the augury inventory-availability
  - Pass in the series/start/end/zone
  - It will return some sort of array of `[date, count]`
- [x] Roll dates up into weeks, and display table

PR #56, opened 2019-10-07
flight inventory table

sandikbarr:
Adds the Inventory Availability Table #30, also expands as described in #32


src/app/campaign/flight/flight.component.ts
```diff
@@ -66,7 +66,7 @@ export class FlightComponent implements OnInit {
 
   ngOnInit() {
     this.flightForm.valueChanges.subscribe(cmp => {
-      this.formStatusChanged(cmp);
+      this.formStatusChanged({...cmp, id: this.flight.id});
```
sandikbarr, line 69: flight is missing the id when it comes out of the Form

src/app/campaign/flight/flight.container.ts
```diff
@@ -13,50 +13,68 @@ import { map } from 'rxjs/operators';
...
     private route: ActivatedRoute,
     private inventoryService: InventoryService,
     private campaignStoreService: CampaignStoreService,
     private router: Router
-  ) {
-    this.route.paramMap.subscribe(params => this.setFlightId(params.get('flightid')));
+  ) {}
+
+  ngOnInit() {
+    this.flightSub = combineLatest(this.route.paramMap, this.campaignStoreService.campaign$).subscribe(([params, campaignState]) => {
```
sandikbarr, line 44: I needed to change this from a) being a nested subscription and b) using `first()` of campaign$. Because `first()` completes after the first update, it broke the unidirectional flow of data. campaign$ is basically root state for this container, and that state is passed down to child components. Child components like the new availability table were not getting updates to the store state because the subscription was cut off by first().

src/app/campaign/availability/availability.component.ts
```diff
@@ -0,0 +1,61 @@
...
+      <div class="row totals">
+        <div class="date">TOTALS</div>
+        <div class="avail">{{ (zone.totals.allocated + zone.totals.availability) | largeNumber }}</div>
+        <div class="goal"></div>
+        <div class="edit"></div>
+      </div>
+    </section>
+  `,
+  styleUrls: ['availability.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
```
pstjean, line 45: 🎉

src/app/campaign/availability/availability.component.ts
```diff
@@ -0,0 +1,61 @@
+import { Component, ChangeDetectionStrategy, Input } from '@angular/core';
+import { Flight, Availability, InventoryZone } from '../../core';
+
+@Component({
+  selector: 'grove-availability',
+  template: `
+    <h2>Impressions</h2>
+    <p *ngIf="flight && (!flight.startAt || !flight.endAt || !flight.set_inventory_uri || !(flight.zones && flight.zones.length))">
```
pstjean, line 8: This logic is doable in the template obviously, but might be easier to manage going forward in the component class.

src/app/campaign/flight/flight.component.ts
pstjean: Yikes, surprised we didn't catch this one earlier.

src/app/campaign/flight/flight.container.ts
```diff
@@ -40,23 +54,28 @@ export class FlightContainerComponent implements OnInit {
...
-      if (state.flights[id]) {
-        this.currentFlightId = id;
-        this.state$.next(state.flights[id]);
-        this.currentInventoryUri$.next(state.flights[id].localFlight.set_inventory_uri);
-      } else {
-        const campaignId = state.remoteCampaign ? state.remoteCampaign.id : 'new';
-        this.router.navigate(['/campaign', campaignId]);
+  setFlightId(id: string, state: CampaignState) {
+    if (state.flights[id]) {
+      if (this.currentFlightId !== id) {
```
pstjean, line 63: I'm guessing this logic is needed because of our decision not to store the temporary flight IDs we create for local flights before they're persisted in the `id` field of a `Flight`. I'm not super convinced this was the right decision, as I'm seeing special case logic like this pop up in a couple of places in my work, and also in this PR. Worth keeping an eye on.

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -55,6 +57,100 @@ export class CampaignStoreService {
...
+            .filter(zone => state.availability[`${flightId}-${zone}`])
+            .map(zone => state.availability[`${flightId}-${zone}`]);
+        // each zone
+        const zoneWeeks = availabilityZones && availabilityZones.map((availability: Availability) => {
+          let weekBeginString: string;
+          let weekEnd: Date;
+          // acc weeks
+          return availability.totals.groups.reduce((acc, day) => {
+            const dayDate = new Date(day.startDate + ' 0:0:0');
+            if (!weekEnd || weekEnd.valueOf() <= dayDate.valueOf()) {
```
pstjean, line 80: I can't readily see how this conditional would ever be `false`.

pstjean (approved):

src/app/campaign/flight/flight.container.ts
sandikbarr: I haven't run across the temporary ids. I wanted to loadAvailability when the flight id param changes. Also I want to set the Observable for availability rollup according to the flight id param. Initially I added the currentFlightId onto the store state and used that in the availability rollup, but I decided I could just leave the current flight id here in the container component for now.

src/app/core/campaign/campaign-store.service.ts
sandikbarr: It breaks the weeks up when they cross over the ending of the previous week. It's true when `weekEnd` is in the previous week, and `dayDate` is day in the following week. It's false when `dayDate` is in the current week. When false, it accumulates onto the acc.totals.groups[weekBeginString]. This could probably be better. We'll be doing more work here I suppose.

pstjean (approved):

PR #59, opened 2019-10-10
Wait for zone options before setting zones

pstjean:
Basically, we were rendering the form before zone options even came through, which meant that state was then updated from the form with 0 of 0 options selected, erasing the previously selected zones. Solution is to wait for options before rendering the form. I’m not really sure that we should even handle this check in the template at this point? Maybe better to pass an unresolved Observable through to the flight form?


sandikbarr (approved):
works 👍

PR #64, opened 2019-10-16
allocation preview

sandikbarr:


src/app/campaign/flight/flight.container.ts
```diff
@@ -81,8 +81,20 @@ export class FlightContainerComponent implements OnInit, OnDestroy {
   }
 
   flightUpdateFromForm({ flight, changed, valid }) {
+    this.flightState$.pipe(
```
sandikbarr: filter to only load the availability when `flight` fields from form have changed as compared to the `localFlight` (different since availability was last loaded)
...maybe belongs in `campaignStore.setFlight()` instead?

src/app/core/inventory/inventory.service.ts
```diff
@@ -38,14 +38,13 @@ export class InventoryService {
...
-  getInventoryAvailability({id, startDate, endDate, zoneName, flightId}): Observable<Availability> {
+  getInventoryAvailability({id, startDate, endDate, zoneName}): Observable<Availability> {
     return this.augury.follow('prx:inventory', {id}).pipe(
       switchMap(inventory => {
         return inventory.follow('prx:availability', {
           startDate,
           endDate,
-          zoneName,
-          flightId
+          zoneName
```
sandikbarr: shouldn't need the fightId to query availability

svevang: Yep,  you should only need that flightId if you want to see the allocation sums for _that_ flight. (instead of sums for all allocations for all flights)

sandikbarr: hmm, maybe I am confused. I took out flightId here because I was working toward being able to see the availability for flights that were not yet created. (I had just given it all the params before without thinking about uncreated flights I guess.) Should I include the flightId for availability when the flight has been created? Is it giving me a sum of all flights otherwise? Or maybe you're just talking allocations, which is my next step. This part is still just working with the availability.

svevang: Yep,  for flights that are created, you would want to include that flightId.  But obviously, for for flights not persisted, then you can't include it.

Sort of a trick of formatting.  In the availability table this means the day is maxed out (no more inventory is available):
Screen Shot 2019-10-16 at 2 49 39 PM

In this case the api response (given flightId) should be something like
```
    {
      "allocated": 5000,
      "availability": 0,
      "date": "2019-04-14"
    },
```
But here is an example where there is 1000 more we could add:
Screen Shot 2019-10-16 at 2 49 45 PM

In this case the api (given flightId) might be something like
```
    {
      "allocated": 5000,
      "availability": 1000,
      "date": "2019-04-14"
    },
```
Or in the case we supply no flight id:
```
    {
      "allocated": <some number of allocations summed, from other flights>,
      "availability": 1000,
      "date": "2019-04-14"
    },
```
In this last case, you'd have to hit the allocation preview endpoint to see what is allocated on that day (hypothetically).

sandikbarr: Ohhhh... I think I get it now, thank you. I was thinking I would be hitting the allocation preview endpoint for all flights, but it's just for the flights that are not created yet.

src/app/campaign/availability/availability.component.ts
```diff
@@ -1,59 +1,92 @@
-import { Component, ChangeDetectionStrategy, Input } from '@angular/core';
+import { Component, ChangeDetectionStrategy, Input, Output, EventEmitter } from '@angular/core';
 import { Flight, Availability, InventoryZone } from '../../core';
 
 @Component({
   selector: 'grove-availability',
   template: `
```
sandikbarr, line 6: This template is getting a bit long, but also this file is mostly template since the data is wrangled in the rolled up observable. 🤷‍♀

src/app/campaign/flight/flight.container.ts
```diff
@@ -43,6 +44,7 @@ import { map } from 'rxjs/operators';
 })
 export class FlightContainerComponent implements OnInit, OnDestroy {
   private currentFlightId: string;
+  private dailyMinimum: number;
```
sandikbarr: I'm not sure I like this dailyMinimum stored here on the container component. It's set from the goal-form and used for requesting the availability preview when the date, series, or zone fields change and the availability preview needs to be reloaded. I could probably store it in the campaign state... 🤔

src/app/core/campaign/campaign.models.ts
```diff
@@ -3,6 +3,7 @@ export interface CampaignState {
   remoteCampaign?: Campaign;
   flights: { [id: string]: FlightState };
   availability?: { [flightZone: string]: Availability };
+  allocationPreview?: { [flightId: string]: { [zone: string]: AllocationPreview } };
```
sandikbarr, line 6: allocationPreview state is shaped for lookup by [flightId][zone], then allocations days are looked up by date.

src/app/core/campaign/campaign.models.ts
```diff
@@ -26,17 +27,35 @@ export interface Campaign {
   set_advertiser_uri: string;
 }
 
-export interface Allocation {
```
sandikbarr, line 29: naming things...so many 'A' words

src/app/campaign/availability/availability.component.spec.ts
```diff
@@ -1,48 +1,52 @@
...
-            {allocated: 0, availability: 1339, startDate: '2019-10-06', endDate: '2019-10-12'},
-            {allocated: 0, availability: 8, startDate: '2019-10-13', endDate: '2019-10-19'},
-            {allocated: 0, availability: 1393, startDate: '2019-10-20', endDate: '2019-10-26'},
-            {allocated: 0, availability: 722, startDate: '2019-10-27', endDate: '2019-10-31'}
-          ]
-        }
-      }];
-      fix.detectChanges();
-    });
+      imports: [SharedModule, NoopAnimationsModule, ReactiveFormsModule, MatFormFieldModule, MatInputModule],
```
sandikbarr, line 18: Turns out I did not have prettier format on save enabled. 🙁

src/app/campaign/availability/availability.component.ts
```diff
@@ -1,59 +1,97 @@
...
+          <div class="edit"><prx-icon name="lock" size="14px" color="secondary"></prx-icon></div>
         </div>
-        <ng-container *ngIf="zoneWeekExpanded[zone.zone + '-' + week.startDate]">
-          <div class="row expand" *ngFor="let day of week.groups">
-            <div class="date"><span>&mdash;</span><span>{{ day.startDate | date:'M/dd' }}</span></div>
-            <div class="avail"><span>&mdash;</span><span>{{ (day.allocated + day.availability) | largeNumber }}</span></div>
-            <div class="goal"></div>
-            <div class="edit"></div>
+        <ng-container *ngFor="let week of zone.totals.groups">
+          <div class="row" [class.row-highlight]="week.allocated + week.availability < week.allocationPreview">
```
sandikbarr, line 23: I'm not sure I understand this row-highlight correctly. I added here what I thought I saw in designs, but it doesn't quite make sense for this. It seems according to this design I should highlight a row if the availability is less than the allocation preview, but I'm not sure why the system would return an allocation preview lower than is available. 
Screen Shot 2019-10-24 at 4 45 21 PM

The criteria in the ticket is this, but Augury already seems to handle these cases because I get an error if daily_min cannot be met instead of a partial result. Unfortunately those error messages don't current come through the hal remote, but we could sort it out.
* Validation errors, if Augury couldn't meet your total/daily_min goals:
  * Highlight offending Total/Daily-Min field in red
  * Change table text to red for offending day/total rows

src/app/campaign/availability/availability.component.ts
```diff
@@ -1,59 +1,97 @@
...
+                  size="6px"
+                  color="primary"
+                >
+                </prx-icon>
+                {{ week.startDate | date: 'M/dd' }}
+              </button>
+            </div>
+            <div class="avail">{{ week.allocated + week.availability | largeNumber }}</div>
+            <div class="goal">
+              {{
```
sandikbarr, line 43: Show the allocation preview if present (even if 0), otherwise the availability allocated should be shown.

The eventual intention here is that the allocation preview will be a temporary state we hold for the flight when looked up (by changing goal, daily min or flight.) Then once allocated upon saving the state, we should clear the allocation preview state for the flight in order to show what has been allocated.

src/app/campaign/availability/goal-form.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    totalGoal: ['', Validators.required],
+    dailyMinimum: ['']
+  });
+  goalFormSub: Subscription;
+
+  constructor(private fb: FormBuilder) {}
+
+  ngOnInit() {
+    this.goalFormSub = this.goalForm.valueChanges
+      .pipe(
```
sandikbarr, line 57: I like that I can debounce right off the reactive form observable.

src/app/campaign/availability/availability.component.ts
pstjean: Sounds like a question for @brandonhundt

src/app/campaign/flight/flight.container.ts
pstjean: Yeah, I'd agree that a non-static value used to make API requests should probably live in state somewhere.

src/app/campaign/flight/flight.container.ts
```diff
@@ -94,11 +97,56 @@ export class FlightContainerComponent implements OnInit, OnDestroy {
...
+          // dates come back as Date but typed string on Flight
+          const flightStartAtDate = flight.startAt && new Date(flight.startAt.valueOf());
+          const flightEndAtDate = flight.endAt && new Date(flight.endAt.valueOf());
+          return (
+            flight.startAt &&
+            flight.endAt &&
+            flightStartAtDate.valueOf() !== flightEndAtDate.valueOf() &&
+            flight.set_inventory_uri &&
+            flight.zones &&
+            flight.zones.length > 0 &&
```
pstjean: This is null checking ...

src/app/campaign/flight/flight.container.ts
```diff
@@ -94,11 +97,56 @@ export class FlightContainerComponent implements OnInit, OnDestroy {
...
+            (flightStartAtDate.valueOf() !== new Date(localFlight.startAt).valueOf() ||
+              flightEndAtDate.valueOf() !== new Date(localFlight.endAt).valueOf() ||
+              flight.set_inventory_uri !== localFlight.set_inventory_uri ||
+              !flight.zones.every(zone => localFlight.zones.indexOf(zone) > -1))
```
pstjean: ... and this is the actual condition we're checking for. The parens do help demarcate this, but I wonder if at this point (~10 LOC) it might be good for comprehension to pull this out into its own function and assign each of these conditional clauses to a variable with a descriptive name, or something like that.

pstjean: Or maybe just assigning two variables right here would do the trick, and pulling this out into its own function is overdoing it.

pstjean (approved):
Suggested a possible change for comprehension. Looks good!

src/app/campaign/availability/availability.component.ts
sandikbarr: #68 has been added to address validation

src/app/campaign/flight/flight.container.ts
sandikbarr: ~I made an attempt at this, but the currentFlightId not being on the state added complications. I had to store a dailyMinimum per flight. Then I tried to set the Observable when the flight id parameter changed, but unfortunately that change doesn't get detected when passed down into the GoalFormComponent. So I'm deciding to leave this.~

sandikbarr: Actually got this working... updating tests

sandikbarr:
Even though it sounds like we'll be getting rid of it, I added that dailyMinimum to the campaign store state so it wasn't stored on the flight form container. Think I've addressed your comments now.

Issue #57, opened 2019-10-08
PRX Angular Material Theme


pstjean:
Right now we're using on of Angular Material's built in themes (the `indigo-pink` theme). We should have colors that match our brand, and [creating an Angular Material theme](https://material.angular.io/guide/theming) is most likely the best way to do that in this context.

brandonhundt:
Noted here: https://docs.google.com/document/d/1WF74k3uzXnjP6w15FKwZzAtU1REmu0ZpQy4JqbfvADc/edit?usp=sharing

PR #65, opened 2019-10-16
Handle deletion and duplication of flights

pstjean:
Uses a `softDeleted` flag to keep flights around in state until a save of the campaign has been initiated by the user.

Unfortunately it looks like Prettier has made some formatting changes that add a bit of noise to the diff. I've tried to separate those into their own commit where I can (i.e. fbf172e5d250869819125b0c44b90f0858f4e9d2)


src/app/campaign/campaign.component.scss
```diff
@@ -29,3 +29,8 @@ mat-drawer {
   text-align: center;
   padding: 12px 0;
 }
+
+.mat-list-base .mat-list-item.deleted-flight {
+  text-decoration: line-through;
+  color: #f00;
```
sandikbarr, line 35: Not sure if you got this red color from the styleguide https://github.com/PRX/styleguide.prx.org/blob/master/projects/ngx-prx-styleguide/src/assets/styles/_variables.scss#L65, but FYI the emphasis colors have typically been reserved for hover/focus/active.

src/app/campaign/campaign.component.ts
```diff
@@ -52,19 +53,19 @@ export class CampaignComponent implements OnInit, OnDestroy {
...
-    this.routeSub = this.route.paramMap.pipe(
-      switchMap((params: ParamMap) => {
-        return this.advertiserService.loadAdvertisers().pipe(
-          map(advertisers => ({params, advertisers}))
-        );
-      })
-    ).subscribe(({params, advertisers}) => {
-      this.campaignStoreService.load(params.get('id'));
-    });
+    this.routeSub = this.route.paramMap
```
sandikbarr, line 56: why the formatting changes? are we not using the same prettier config?

src/app/campaign/flight/flight.component.html
```diff
@@ -35,4 +35,17 @@
...
     </mat-form-field>
   </form>
+  <div class="flight-controls-container">
+    <button
+      (click)="toggleDeleteFlight()"
+      mat-fab
+      [color]="softDeleted ? 'primary' : 'warn'"
+      [attr.aria-label]="softDeleted ? 'Restore flight' : 'Delete flight'"
+    >
+      <mat-icon>{{ softDeleted ? 'restore_from_trash' : 'delete' }}</mat-icon>
```
sandikbarr, line 45: we're actively using several icon implementations, the prx-icon in the styleguide built by Rick, the global icon styles, and now material icons, yay more tech debt

src/app/campaign/flight/flight.component.ts
```diff
@@ -81,4 +96,12 @@ export class FlightComponent implements OnInit {
   updateFlightForm({ startAt, endAt, ...restOfFlight }: Flight) {
     this.flightForm.reset({ startAt: new Date(startAt), endAt: new Date(endAt), ...restOfFlight }, { emitEvent: false });
   }
+
```
sandikbarr, line 99: 🙃  _smiles at reverse naming_
Typically, Angular style would have you name the functions similarly but prefixed with "on". So `onFlightDuplicate()` would call `this.flightDuplicate.emit(this.flight)` or `onDuplicateFlight() with `duplicateFlight.emit()`. The convention is to use "on" prefix with the function and without the prefix in the template/Output.

src/app/campaign/flight/flight.container.ts
```diff
@@ -85,4 +98,24 @@ export class FlightContainerComponent implements OnInit, OnDestroy {
     this.campaignStoreService.loadAvailability(flight);
     this.currentInventoryUri$.next(flight.set_inventory_uri);
   }
+
+  flightDuplicate(flight: Flight) {
+    const localFlight: Flight = { ...flight, name: `${flight.name} (Copy)` };
+    this.campaignStoreService.campaignFirst$.subscribe(state => {
+      const flightId = Date.now();
```
sandikbarr, line 105: should we encapsulate this temp flightId generation in the campaignStoreService?

src/app/campaign/flight/flight.container.ts
```diff
@@ -85,4 +98,24 @@ export class FlightContainerComponent implements OnInit, OnDestroy {
...
+
+      const campaignId = state.remoteCampaign ? state.remoteCampaign.id : 'new';
+      this.router.navigate(['/campaign', campaignId, 'flight', flightId]);
+    });
+  }
+
+  flightDeleteToggle() {
+    this.campaignStoreService.campaignFirst$.subscribe(state => {
+      const currentState = state.flights[this.currentFlightId];
+      const newState = { ...currentState, changed: true, softDeleted: !!!currentState.softDeleted };
```
sandikbarr: triple `!!!`? is this making sure it's extra soft deleted?

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -62,73 +61,85 @@ export class CampaignStoreService {
...
-          }, {
-            zone: availability.zone,
-            totals: {
-              startDate: availability.totals.startDate,
-              endDate: availability.totals.endDate,
-              allocated: 0,
-              availability: 0,
-              groups: {}
-            }
+        const zoneWeeks =
```
sandikbarr, line 74: whitespace prettier config?

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -139,34 +150,47 @@ export class CampaignStoreService {
...
   }
 
   storeCampaign(): Observable<CampaignStateChanges> {
     const saving = this.putCampaign().pipe(
       switchMap(campaignChanges => {
+        const { flights } = campaignChanges.state;
+        const deletedFlightIds = Object.keys(flights).filter(fId => flights[fId].softDeleted);
+        deletedFlightIds.forEach(fId => {
+          if (flights[fId].remoteFlight) {
+            this.campaignService.deleteFlight(fId).subscribe();
```
sandikbarr: code smell to subscribe inside an observable stream, typically you would return the observable and use the result in the final subscribe

sandikbarr (commented):
some initial comments on first pass..

sandikbarr:
Screen Shot 2019-10-17 at 9 56 29 AM
The datepicker buttons look interesting on delete/disable 😬

src/app/campaign/campaign.component.scss
pstjean: This was a @brandonhundt suggestion. I think this will probably end up being changed when #57 gets done (along with the horrible global style breakage you'll see on disable form fields and quite a few other things).

src/app/campaign/campaign.component.ts
pstjean: We should be, since the `.prettierrc` is committed to Git. A couple of possibilities (assuming you're using VSCode):

- Your formatter isn't always running. I've enabled format on save in the global settings, which does the trick.
Screen Shot 2019-10-17 at 10 59 49 AM

- You don't have the Prettier extension installed, or have a formatter other than Prettier set as the default. You can check/change this by opening up the command palette within a TS file and looking at the options under "Format Document With ..."

Screen Shot 2019-10-17 at 11 10 25 AM

src/app/campaign/flight/flight.component.html
pstjean: 🙁 ... yay. Again hoping for clarity after #57.

src/app/campaign/flight/flight.component.ts
pstjean: Good call! I knew the reverse naming was silly while I was doing it, couldn't come up with anything better at the time.

src/app/campaign/flight/flight.container.ts
pstjean: Ha! As an optional property, it's sometimes `undefined`, so for some reason I thought it needed to be coerced this way. It doesn't.

src/app/core/campaign/campaign-store.service.ts
pstjean: Must be.

pstjean:
> Screen Shot 2019-10-17 at 9 56 29 AM
> The datepicker buttons look interesting on delete/disable 😬

Yeah this is a mess. Global `ngx-prx-styleguide` styles. I'm starting to think #57 should be an epic.

sandikbarr (approved):
Tested this out now as well, and it all works well, so I'm ready to approve. 😄

src/app/core/campaign/campaign-store.service.ts
pstjean: Good shout, meant to circle back to this.

sandikbarr (approved):
💯

sandikbarr (approved):

PR #72, opened 2019-10-29
fixes dailyMinimum undefined error

sandikbarr:
dailyMinimum will likely be removed anyway, but I added dailyMinimum to the campaign store state since it came up in review of #64, then I spotted an undefined error with it.


cavis (approved):
👍 makes sense to me

PR #73, opened 2019-10-29
Introducing 'prx-theme.scss' 

brandonhundt:
This PR includes a new file `prx-theme.scss` that includes variables for the entire PRX app color palette. See this for more detail: https://material.angular.io/guide/theming

To test, I'd recommend referencing this in your component and changing a few color variables to see that it works properly. Because it might not. And if not, I'll need some help getting this working.

Thanks!

### DIFF
--- a/src/app/campaign-list/card/campaign-card.component.scss
+++ b/src/app/campaign-list/card/campaign-card.component.scss
@@ -39,5 +39,5 @@ h3 {
   background-color: #ff9600;
 }
 .sold {
-  background-color: #17C9D3;
+  background-color: #17c9d3;
 }
--- a/src/app/campaign-list/card/campaign-card.component.scss
+++ b/src/app/campaign-list/card/campaign-card.component.scss
@@ -1,10 +1,12 @@
+@import '~src/sass/variables';
+
 :host {
   display: block;
   background-color: white;
 }
 .header {
   height: 3px;
-  background-color: #a3a3a3;
+  background-color: map-get($prx-gray, 500);
 }
 section {
   padding: 10px;
@@ -12,7 +14,7 @@ section {
 }
 section.loading {
   position: relative;
-  background-color: #ddd;
+  background-color: map-get($prx-gray, 300);
 }
 h3 {
   font-weight: 700;
@@ -21,23 +23,23 @@ h3 {
 }
 .status {
   padding: 2px 4px;
-  color: #000;
+  color: map-get($prx-gray, 900);
 }
 .draft {
-  background-color: #b3d9ef;
+  background-color: map-get($prx-blue, 200);
 }
 .approved {
-  background-color: #23c763;
+  background-color: map-get($prx-green, 500);
 }
 .paused {
-  background-color: #ffc400;
+  background-color: map-get($prx-yellow, 500);
 }
 .canceled {
-  background-color: #e04524;
+  background-color: map-get($prx-red, 500);
 }
 .hold {
-  background-color: #ff9600;
+  background-color: map-get($prx-orange, 500);
 }
 .sold {
-  background-color: #17c9d3;
+  background-color: map-get($prx-aqua, 500);
 }
--- a/src/app/campaign-list/card/campaign-card.component.scss
+++ b/src/app/campaign-list/card/campaign-card.component.scss
@@ -1,12 +1,13 @@
-@import '~src/sass/variables';
+@import '~src/sass/colors';
+@import '~@angular/material/theming';
 
 :host {
   display: block;
-  background-color: white;
+  background-color: mat-color($prx-gray, 100);
 }
 .header {
   height: 3px;
-  background-color: map-get($prx-gray, 500);
+  background-color: mat-color($prx-gray, 500);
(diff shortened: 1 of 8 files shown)


src/prx-theme.scss
```diff
@@ -0,0 +1,185 @@
...
+  )
+);
+
+$prx-primary: mat-palette($prx-blue, 500, 200, 800);
+$prx-neutral: mat-palette($prx-gray, 500, 200, 800);
+$prx-accent: mat-palette($prx-orange, 500, 200, 800);
+
+$prx-warn: mat-palette($prx-red);
+
+$prx: mat-light-theme(
```
sandikbarr: seeing another error that seems like the `mat-light-theme` only takes three arguments: `mat-light-theme($app-primary, $app-accent, $app-warn)`

sandikbarr:
added WIP to title so it won't pester in slack since this PR is not ready

src/prx-theme.scss
```diff
@@ -0,0 +1,174 @@
+@import '~@angular/material/theming';
+
+@include mat-core();
+
```
sandikbarr: One thing I'm seeing strongly suggested here https://medium.com/@aleixsuau/how-to-use-angular-material-2-sass-variables-in-your-components-76ce0203f126 is defining the variables in a separate file so that we don't have to import mat-core and angular-material-theme every time we import prx-theme.scss from other files to access the variables.

brandonhundt: Great. I'll take a look at that article. Thanks!!!

src/prx-theme.scss
```diff
@@ -0,0 +1,174 @@
...
+    400: $dark-primary-text,
+    500: $dark-primary-text,
+    600: $light-primary-text,
+    700: $light-primary-text,
+    800: $light-primary-text,
+    900: $light-primary-text
+  )
+);
+
+$prx-primary: mat-palette($prx-blue, 500, 200, 800);
```
sandikbarr: not entirely sure I'm doing it right, but trying to use this $prx-primary variable in a component's scss after importing this theme file results in an invalid CSS error:
```
$result: map_merge($result, (
grove_1  |             ^
grove_1  |       (100: #e6f4f9, 200: #b3d9ef, 300: #75bbe1, 400: #3f94c6, 500: #0089bd, 600: #0072a3, 700: #004e75, 800: #003a57, 900: #002533, contrast: (100: rgba(0, 0, 0, 0.87), 200: rgba(0, 0, 0, 0.87), 300: rgba(0, 0, 0, 0.87), 400: rgba(0, 0, 0, 0.87), 500: white, 600: white, 700: white, 800: white, 900: white), default: #0089bd, lighter: #b3d9ef, darker: #003a57, default-contrast: white, lighter-contrast: rgba(0, 0, 0, 0.87), darker-contrast: white, "100-contrast": rgba(0, 0, 0, 0.87), "200-contrast": rgba(0, 0, 0, 0.87), "300-contrast": rgba(0, 0, 0, 0.87), "400-contrast": rgba(0, 0, 0, 0.87), "500-contrast": white, "600-contrast": white, "700-contrast": white, "800-contrast": white, "900-contrast": white, "contrast-contrast": null) isn't a valid CSS value.
```

sandikbarr: This was user error. I was missing this critical piece of information in the `mat-color()` function https://github.com/angular/components/blob/master/guides/theming-your-components.md#note-using-the-mat-color-function-to-extract-colors-from-a-palette

src/app/campaign-list/card/campaign-card.component.scss
```diff
@@ -21,23 +23,23 @@ h3 {
 }
 .status {
   padding: 2px 4px;
-  color: #000;
+  color: map-get($prx-gray, 900);
```
sandikbarr: making this off black rather than true black makes the contrast issue we have on these cards worse. True black had a better contrast ratio than white with all these colors, but it still wasn't a11y 4.5:1 contrast compliant.

src/app/campaign-list/card/campaign-card.component.scss
```diff
@@ -1,18 +1,20 @@
+@import '~src/sass/variables';
+
 :host {
   display: block;
   background-color: white;
```
sandikbarr: should there be a $prx-white or some kind of grey?

pstjean: Switched over to the lightest gray, but we could define a PRX white (although I'd lean towards just using `black` or `white` if we're just going to define a variable that is `#000000` or `#ffffff`)

src/app/campaign-list/card/campaign-card.component.scss
pstjean: Ah, this is a good use of the palette's `<hue>-contrast`, so I've switched over to that. It _looks_ like the `@angular/material/_theming.scss` file is just mixin, function and variable definitions, so it should be no problem to import that wherever we need to use those function calls (as far as I can tell).

pstjean (approved):
👍

PR #74, opened 2019-10-30
Create campaign

pstjean:
Campaign creation wasn't redirecting properly due to a bug when creating a campaign without flights.


cavis (approved):
👍

PR #100, opened 2019-12-27
dashboard flight table

sandikbarr:
Adds a flight table to the dashboard using material tabs, table and paging. Flight table is filtered against the augury flight filtering and facet API.

The DashboardService (previously the CampaignListService) holds the route params and the campaign and dashboard lists state. The CampaignListComponent and the FlightTableComponent listen on the activated route for route changes and call the service to set the route params and load either the campaign or flight list. When filters are selected from the dropdowns or text entry, the route is changed causing a new filtered campaign or flight list to be loaded.


src/app/app.routing.module.ts
```diff
@@ -4,18 +4,18 @@ import { Routes, RouterModule } from '@angular/router';
 import { AuthGuard, UnauthGuard } from 'ngx-prx-styleguide';
 
 import { AuthorizationComponent } from './authorization/authorization.component';
-import { HomeComponent } from './home/home.component';
+import { DashboardComponent } from './dashboard/dashboard.component';
```
sandikbarr, line 7: Did some refactor/rename/move of some of these files from the original campaign list on the dashboard to fit the flight table within it.

src/app/dashboard/campaign-card/campaign-flight-dates.pipe.ts
```diff
@@ -0,0 +1,34 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Flight } from '../dashboard.service';
```
sandikbarr, line 2: This file was just moved, no changes

svevang (approved):
Looks good!

PR #101, opened 2020-01-02
shows flight allocation when saved

sandikbarr:
* when flight is changed/unsaved the allocation preview is shown
* otherwise when flight is saved, availability is reloaded and the saved allocation is shown
  * preview numbers are shown in green (`#178240`) with an asterisk*

### DIFF
--- a/src/app/campaign/availability/availability.component.scss
+++ b/src/app/campaign/availability/availability.component.scss
@@ -1,3 +1,6 @@
+@import '~src/sass/colors';
+@import '~@angular/material/theming';
+
 :host {
   margin: 20px 10px 20px 20px;
 }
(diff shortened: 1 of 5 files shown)


svevang (approved):
Looks good! Transition from preview and back works great.

I discovered a pretty good issue on augury that was causing 500's as I tested this branch. This prevents and PUT or POST from attempting to save allocations:
https://github.com/PRX/augury.prx.org/pull/155/files#diff-0cda254475c0c7fff7b657c5b34b7609R14

Also after some digging, found an unrelated issue with inventory counts, which should fix up some strange behavior in available inventory and previews: https://github.com/PRX/grove.prx.org/issues/102

PR #110, opened 2020-01-14
Fix campaign owner field

sandikbarr:
Fixes the campaign owner field to load the campaign after the accounts have been loaded.

Also a workaround on the issue currently on campaign/flights zoom where the embedded flights list is not paged, and Hal is not handling it. I removed the zoom on flights there to get around the error.

### DIFF
--- a/src/app/campaign/campaign.component.ts
+++ b/src/app/campaign/campaign.component.ts
@@ -2,7 +2,7 @@ import { Component, ChangeDetectionStrategy, OnInit, OnDestroy } from '@angular/
 import { ActivatedRoute, ParamMap, Router } from '@angular/router';
 import { Observable, Subscription } from 'rxjs';
 import { map, switchMap } from 'rxjs/operators';
-import { CampaignStoreService, AdvertiserService } from '../core';
+import { CampaignStoreService, AdvertiserService, AccountService } from '../core';
 import { ToastrService } from 'ngx-prx-styleguide';
 
 @Component({
@@ -49,6 +49,7 @@ export class CampaignComponent implements OnInit, OnDestroy {
     private router: Router,
     private toastr: ToastrService,
     public campaignStoreService: CampaignStoreService,
+    private accountService: AccountService,
     private advertiserService: AdvertiserService
   ) {}
 
@@ -57,9 +58,12 @@ export class CampaignComponent implements OnInit, OnDestroy {
       .pipe(
         switchMap((params: ParamMap) => {
           return this.advertiserService.loadAdvertisers().pipe(map(advertisers => ({ params, advertisers })));
+        }),
+        switchMap(({ params, advertisers }) => {
+          return this.accountService.loadAccounts().pipe(map(accounts => ({ params, advertisers, accounts })));
         })
       )
-      .subscribe(({ params, advertisers }) => {
+      .subscribe(({ params, advertisers, accounts }) => {
         this.campaignStoreService.load(params.get('id'));
       });
     this.campaignFlights$ = this.campaignStoreService.flights$.pipe(
--- a/src/app/campaign/flight/flight.container.spec.ts
+++ b/src/app/campaign/flight/flight.container.spec.ts
@@ -94,7 +94,8 @@ describe('FlightContainerComponent', () => {
     });
   });
 
-  it('redirects to campaign if the flight does not exist', () => {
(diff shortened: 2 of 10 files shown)


src/app/core/account/account.service.ts
```diff
@@ -12,15 +12,27 @@ export interface Account {
 
 @Injectable()
 export class AccountService {
+  // tslint:disable-next-line: variable-name
+  _accounts: BehaviorSubject<Account[]> = new BehaviorSubject([]);
+
+  get accounts(): Observable<Account[]> {
+    return this._accounts.asObservable();
```
sandikbarr, line 19: I turned accounts into a BehaviorSubject so the accounts are loaded by the container component and other components can just subscribe to the list of accounts.

src/app/core/campaign/campaign.service.ts
```diff
@@ -13,8 +13,11 @@ export class CampaignService {
   constructor(private augury: AuguryService) {}
 
   getCampaign(id: number | string): Observable<CampaignState> {
-    return this.augury.follow('prx:campaign', { id, zoom: 'prx:flights' }).pipe(
-      switchMap(doc => doc.followItems('prx:flights').pipe(map(flightDocs => ({ doc, flightDocs })))),
```
sandikbarr: There's an error on staging right now due to a change in the API. This was loading the campaign with flights embedded with zoom. However, that embedded list of flights is unexpectedly an unpaged list and causes the Hal support to throw an error. So I changed it to load the flights separately. It now assumes that campaign/flights is a paged list by default, so it loads the paged list of flights to get the total amount, then loads the full list of flights using total as the per param.

src/app/campaign/campaign.component.ts
```diff
@@ -57,9 +58,12 @@ export class CampaignComponent implements OnInit, OnDestroy {
       .pipe(
         switchMap((params: ParamMap) => {
           return this.advertiserService.loadAdvertisers().pipe(map(advertisers => ({ params, advertisers })));
+        }),
```
sandikbarr, line 61: This is the bug fix for accounts. If the campaign loaded before accounts, and the reactive form was bound when accounts had not yet loaded, accounts would not be populated. This makes the campaign load dependent on loading accounts first (same as we're doing for advertisers.)

src/app/core/campaign/campaign.service.ts
```diff
@@ -14,7 +14,15 @@ export class CampaignService {
 
   getCampaign(id: number | string): Observable<CampaignState> {
     return this.augury.follow('prx:campaign', { id, zoom: 'prx:flights' }).pipe(
-      switchMap(doc => doc.followItems('prx:flights').pipe(map(flightDocs => ({ doc, flightDocs })))),
+      switchMap(doc => doc.follow('prx:flights').pipe(map(flights => ({ doc, flights })))),
+      switchMap(({ doc, flights }: { doc: HalDoc; flights: HalDoc }) => {
```
sandikbarr, line 18: Checks if total is greater than count/page size and if so, selects all flights with `{per: total}` param; otherwise, uses flights embedded into campaign response.

src/app/core/account/account.service.ts
cavis: +1

src/app/core/campaign/campaign.service.ts
```diff
@@ -14,7 +14,15 @@ export class CampaignService {
...
+        // if total is greater than count, request all flights
+        let params: any;
+        if (+flights['total'] > +flights['count']) {
+          params = { per: +flights['total'] };
+        }
+        return doc.followItems('prx:flights', params).pipe(map(flightDocs => ({ doc, flightDocs })));
```
cavis, line 24: So this should work whether or not `prx:flights` comes back paged or not, right?  (This would be a good addition to HalObservable itself, someday).

src/app/campaign/flight/flight.container.spec.ts
```diff
@@ -94,7 +94,8 @@ describe('FlightContainerComponent', () => {
     });
   });
 
-  it('redirects to campaign if the flight does not exist', () => {
+  // TODO: no longer does this because it caused a premature navigation bug, should we create a better way to detect it?
```
cavis, line 97: Yeah, as long as there's a TODO on this, I'm fine with re-introducing this bug.  It's such an edge case anyways.

cavis (approved):
👍

PR #119, opened 2020-02-05
API changes

svevang:
Changes the availability attribute name in the availability response.

To be merged concurrently with https://github.com/PRX/augury.prx.org/pull/168


cavis (approved):
👍

PR #120, opened 2020-02-05
fixes loading flight allocation preview on form change

sandikbarr:
When opening a new flight form, I found there is an async issue between updating the state and the form updates, so I added a change to make sure the flight ids match before loading the availability and allocation preview on form change (in addition to all the other criteria for valid fields.)


cavis (approved):
👍

Issue #111, opened 2020-01-16
use material theme


sandikbarr:
clean up Grove styles to use the material theming

PR #123, opened 2020-02-12
#111: wire up material theme colors

rpeterman-gp:
- add sass functions for theming
- replace static colors with theme variables or theme map prop
- adjustment made to campaign form layout
- flight table paginator should not scroll with overflowed cells

### To Review

- [ ] Checkout branch.
- [ ] Run `yarn`.
- [ ] Run `docker-compose build`.
- [ ] Run `docker-compose up`.
- [ ] Start `http-proxy`.
- [ ] Go to [grove.prx.docker](https://grove.prx.docker).
- [ ] Ensure theme coloring is as expected.
- [ ] Edit `/sass/_variables.scss` and change the function used on line 13 to `mat-dark-theme`.
- [ ] Ensure dark theme is applied as expected (this is to demonstrate the theme variables are wired up correctly.)
- [ ] Go to Flights table and adjust screen so the table can scroll horizontally.
- [ ] Ensure paginator stays in place as you scroll the table.
- [ ] Go to a campaign and note the fields layout with horizontal space between fields when they are in two columns.

### DIFF
--- a/src/app/dashboard/flight-table/flight-table.component.scss
+++ b/src/app/dashboard/flight-table/flight-table.component.scss
@@ -23,11 +23,11 @@ $cell-padding: $cell-padding-v $cell-padding-h;
 }
 
 .mat-table-sticky:first-child {
-  border-right: 1px solid mat-color($prx-gray, 300);
+  border-right: 1px solid prx-theme-foreground(divider);
 }
 
 .mat-header-cell {
-  background: mat-color($prx-gray, 300);
+  background: prx-theme-background(table-header);
   border: white;
   padding: $cell-padding;
   line-height: $row-height;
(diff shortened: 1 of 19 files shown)


src/app/dashboard/flight-table/flight-table.component.scss
```diff
@@ -82,3 +82,8 @@ $cell-padding: $cell-padding-v $cell-padding-h;
     color: $warn-contrast;
   }
 }
+
```
sandikbarr, line 85: what's this change for?

sandikbarr (changes requested):
looks really great, just one question on that sticky paginator change. thanks for doing this!

src/app/dashboard/flight-table/flight-table.component.scss
rpeterman-gp: I noticed when the table scrolled horizontally, the paginator would also move. Currently it scrolls almost halfway across the container, but if the table were to contain more columns or wider column content, could end up scrolling to a point some  of the buttons could be come inaccessible. This change makes the paginator stay in one place when the table is scrolled horizontally.

sandikbarr: good demo, makes sense. thanks!

sandikbarr (approved):
looks great!

PR #136, opened 2020-02-20
ngrx campaign form store

sandikbarr:


src/app/campaign/campaign.component.spec.ts
```diff
@@ -1,22 +1,123 @@
-import { CampaignComponent } from './campaign.component';
-import { CampaignStoreService, CampaignState, FlightState, AccountService, AdvertiserService } from '../core';
+import { TestBed, async, ComponentFixture } from '@angular/core/testing';
```
sandikbarr, line 1: This component spec didn't have a proper TestBed, so I worked one up and set up all the dependencies.

src/app/campaign/campaign.component.ts
```diff
@@ -77,19 +82,26 @@ export class CampaignComponent implements OnInit, OnDestroy {
       )
       .subscribe(({ params, advertisers, accounts }) => {
         this.campaignStoreService.load(params.get('id'));
```
sandikbarr, line 84: This also still loads the campaign in the campaign store. I think we can live with that for now while we transition.

src/app/campaign/campaign.component.ts
```diff
@@ -53,19 +56,21 @@ export interface LocalFlightState {
...
+    public store: Store<any>,
     private accountService: AccountService,
     private advertiserService: AdvertiserService
   ) {}
 
   ngOnInit() {
+    this.campaignSaving$ = this.store.pipe(select(selectCampaignSaving));
     this.routeSub = this.route.paramMap
       .pipe(
         switchMap((params: ParamMap) => {
```
sandikbarr, line 76: 🤔 We'll probably want advertisers, accounts(, and inventories/series) to be NgRx state as well. Advertisers and Accounts are currently RxJS BehaviorSubject stores.

This is the fragile bit with reactive forms where we make sure to load the advertisers and accounts lists ahead of the campaign so the form has the lists to populate the dropdowns and match a value in order to set the form component value. Yuck.

src/app/campaign/campaign.component.ts
```diff
@@ -101,19 +113,26 @@ export class CampaignComponent implements OnInit, OnDestroy {
...
-      const flightId = this.router.url.split('/flight/').pop();
-      if (this.router.url.includes('/flight/') && !changes.flights[flightId]) {
-        this.router.navigate(['/campaign', changes.id]);
-      } else if (this.router.url.includes('/flight/') && flightId !== changes.flights[flightId]) {
-        this.router.navigate(['/campaign', changes.id, 'flight', changes.flights[flightId]]);
-      } else if (changes.prevId !== changes.id) {
-        this.router.navigate(['/campaign', changes.id]);
+      // TODO: when on a flight with a new campaign, don't have the new campaign.id
+      //  so only route to flight if campaign already existed,
+      //  otherwise just let campaign.effects route to the new campaign for now
```
sandikbarr, line 126: TODO further info: I moved the new campaign routing to campaign.effects, but I couldn't move this flight routing yet because effects doesn't handle flights yet. The campaign store no longer has information on the new campaign id because it no longer actually creates/saves the campaign, but the campaign store is still handling the flights. Previously, this was handling routing back to a new flight on a new campaign, but for now the effects routes to the campaign when you create both a flight and a campaign together. Will fix when adding flight handling to effects.

src/app/campaign/form/campaign-form.component.ts
```diff
@@ -102,18 +102,24 @@ export class CampaignFormComponent implements OnInit {
...
-      ...(repName && {repName}),
-      ...(notes && {notes}),
-      ...(set_account_uri && {set_account_uri}),
-      ...(findAdvertiserByURI && {set_advertiser_uri: findAdvertiserByURI.set_advertiser_uri})
-    }, {emitEvent: false, onlySelf: true});
+  updateCampaignForm(campaign: Campaign) {
+    const { name, type, status, repName, notes, set_account_uri, set_advertiser_uri } = campaign;
+    const findAdvertiserByURI =
+      set_advertiser_uri && this.advertisers && this.advertisers.find(adv => adv.set_advertiser_uri === set_advertiser_uri);
+    // only set fields that are present, but allow fields to be explicitly set to null or empty string (new campaign)
```
sandikbarr, line 109: changed to allow the campaign form to be reset for a new campaign with null/empty fields

src/app/campaign/form/campaign-form.container.spec.ts
```diff
@@ -1,21 +1,63 @@
...
+  let de: DebugElement;
+  let el: HTMLElement;
+  const campaignStoreService: CampaignStoreService = { localCampaign$: of({}), setCampaign: jest.fn(() => of({})) } as any;
 
-  beforeEach(() => {
-    accountService = { loadAccounts: jest.fn(() => of([])) } as any;
-    advertiserService = new AdvertiserServiceMock(new MockHalService()) as any;
-    campaignStoreService = { localCampaign$: of({}), setCampaign: jest.fn(() => of({})) } as any;
-    component = new CampaignFormContainerComponent(campaignStoreService, accountService, advertiserService);
-  });
```
sandikbarr, line 18: added TestBed

src/app/campaign/form/campaign-form.container.ts
```diff
@@ -16,20 +19,26 @@ import { withLatestFrom } from 'rxjs/operators';
...
   ) {}
 
+  ngOnInit() {
+    this.campaign$ = this.store.pipe(select(selectLocalCampaign));
+  }
+
   campaignUpdateFromForm(newState: { campaign: Campaign; changed: boolean; valid: boolean }) {
     const { campaign, changed, valid } = newState;
     this.campaignStoreService.setCampaign({ localCampaign: campaign, changed, valid });
+    this.store.dispatch(new CampaignFormUpdate({ campaign, changed, valid }));
```
sandikbarr, line 41: still doing both campaign store and NgRx store form updates, but campaign store will be removed eventually

src/app/campaign/store/reducers/campaign.reducer.ts
```diff
@@ -0,0 +1,103 @@
+import { CampaignActions } from '../actions/';
+import { CampaignActionTypes } from '../actions/campaign.action.types';
+import { HalDoc } from 'ngx-prx-styleguide';
+
```
sandikbarr: TODO: have a few too many Campaign interfaces declared in this app. There is still one in core/campaign.models because it's used by the campaign.service, but that campaign.service (and the rest of those core services) are only used by this campaign module (I think.) Probably will move them to the feature module where it makes sense to keep things more localized.

src/app/core/advertiser/advertiser.service.ts
```diff
@@ -1,6 +1,6 @@
 import { Injectable } from '@angular/core';
```
sandikbarr, line 1: whitespace

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -367,13 +367,17 @@ export class CampaignStoreService {
 
   private putCampaign(): Observable<{ prevId: number; state: CampaignState }> {
     return this.campaignFirst$.pipe(
-      switchMap(state => {
-        return this.campaignService.putCampaign(state).pipe(
-          map(newState => {
-            return { prevId: state.remoteCampaign ? state.remoteCampaign.id : null, state: newState };
-          })
-        );
-      })
```
sandikbarr, line 376: fake the campaign put/post by just returning the campaign from the state as the new/updated campaign. So on new campaigns, the changes won't contain the new campaign id. The campaign store eventually gets the campaign id when it reloads the campaign on route change.

src/app/core/campaign/campaign.service.spec.ts
```diff
@@ -49,6 +51,11 @@ describe('CampaignService', () => {
       valid: true,
       flights: { 9: flightStateFixture }
     };
+    augury = new MockHalService();
+    campaignDoc = augury.mock('prx:campaign', campaignFixture);
+    store = of({ campaignState: { ...campaignStateFixture, doc: campaignDoc } }) as any;
```
sandikbarr, line 56: Might want to create a TestBed for this spec at some point.
I didn't want to mock up the entire store setup here to include the campaign doc selector, so I mocked the store just as an Observable that returns state.

src/app/core/campaign/campaign.service.ts
```diff
@@ -1,16 +1,24 @@
...
 import { HalDoc } from 'ngx-prx-styleguide';
 import { AuguryService } from '../augury.service';
 import { CampaignState, Campaign, Flight, FlightState } from './campaign.models';
+import { selectCampaignDoc } from '../../campaign/store/selectors';
 
 @Injectable()
 export class CampaignService {
-  private campaignDoc: HalDoc;
-  private flightDocs: { [id: number]: HalDoc };
+  campaignDoc$: Observable<HalDoc>;
```
sandikbarr, line 12: This service uses the HalDoc to later make updates, delete/destroy, or create flights on a campaign. Instead of keeping the HalDoc here in the service and relying on it having been set in order to create flights or update the campaign, it's been moved to the store and gets saved each campaign get/put/post. This way, when creating flights on a new campaign, we can wait on that HalDoc Observable to have been set in order to create the flights.

src/app/core/campaign/campaign.service.ts
```diff
@@ -38,38 +45,70 @@ export class CampaignService {
...
       return this.augury.root.pipe(
         switchMap(rootDoc => rootDoc.create('prx:campaign', {}, state.localCampaign)),
         map(doc => {
-          this.campaignDoc = doc;
-          return this.docToCampaign(doc);
+          return this.docToCampaignState(doc);
         })
       );
     }
   }
```
sandikbarr, line 61: loadCampaignZoomFlights basically does the same things as getFlights, but it returns Campaign and Flight models instead of State models. I want to keep State model awareness inside the store and have everything else just use the plain models. updateCampaign/createCampaign also duplicate functionality in putCampaign so we're just working with the plain models outside the store.

src/app/core/campaign/campaign.service.ts
```diff
@@ -79,7 +118,20 @@ export class CampaignService {
     return this.flightDocs[flightId].destroy();
   }
 
-  docToCampaign(doc: HalDoc): CampaignState {
```
sandikbarr, line 82: ...eventually won't need both of these either

src/app/store/reducers/index.ts
```diff
@@ -0,0 +1,8 @@
+import { ActionReducer, ActionReducerMap, createFeatureSelector, createSelector, MetaReducer } from '@ngrx/store';
+import { environment } from '../../../environments/environment';
+
+export interface State {}
```
sandikbarr, line 4: Setup for the global app state (as opposed to feature state.) Seems like most of what we're doing belongs in feature state, not sure yet what we'll be putting in global state (besides the route from router-store.)

src/app/store/router-store/custom-router-serializer.ts
```diff
@@ -0,0 +1,32 @@
...
+  url: string;
+  params: Params;
+  queryParams: Params;
+}
+
+export class CustomRouterSerializer implements RouterStateSerializer<RouterStateUrl> {
+  serialize(routerState: RouterStateSnapshot): RouterStateUrl {
+    let route = routerState.root;
+    let params = {};
+
```
sandikbarr, line 14: This serializer runs on all the routing events and allows us to get the params from all the route segments for digesting into the router-store.

src/app/core/campaign/campaign.service.ts
```diff
@@ -79,7 +118,20 @@ export class CampaignService {
...
+  docToCampaign(doc: HalDoc): Campaign {
```
sandikbarr: move to reducer

src/app/campaign/campaign.component.spec.ts
cavis: +100.  I think we started out with only 1 or 2 "mocked" constructor args.  But ... time makes fools of us all.  Became way too many injected things to be manually mocked.

src/app/campaign/campaign.component.ts
cavis: I've actually been thinking we could just do a better job handling out-of-order loading.

Ex: if the list flight loads before the list of advertisers/accounts ... then set the dropdown options = your currently selected option (and some dummy "Display" value).  Could even attempt to wire that into NgRx somehow?  Then when the options-list comes in, change the dropdown options again.

src/app/campaign/campaign.component.ts
cavis: Makes sense.  Not worth spending a ton of time on these edge-cases pre router-store.

src/app/campaign/store/actions/campaign.action.creator.ts
```diff
@@ -0,0 +1,56 @@
+import { Action } from '@ngrx/store';
```
cavis, line 1: I like that these are nested/specific to their modules.  (The "campaign form" module in this case ... though we just called it Campaign")

src/app/campaign/store/reducers/campaign.reducer.ts
cavis: We also talked about maybe just sending the HalDoc (or basic objects?) all the way out to the forms and letting them deal directly with the JSON our API returns.  Not sure I know enough to know why that's a bad idea.

Anyways ... not in this PR for sure, but might think about going forward.

src/app/core/advertiser/advertiser.service.ts
cavis: Gotta get me one of them fancy lint-on-save editors.  I actually did finally get eslint autofixing in Atom.  Just haven't tried a typescript project yet.

src/app/core/campaign/campaign.service.ts
cavis: +1 good call.

src/app/core/campaign/campaign.service.ts
cavis: +2 also a good call.  Encapsulate that state!

cavis (approved):
:+99: things are still working!  Looks good to me.  Seems like this will really clean up our components.

src/app/campaign/campaign.component.ts
sandikbarr: I could see that working pretty easily with NgRx yeah.

src/app/campaign/store/reducers/campaign.reducer.ts
sandikbarr: I do just about have a change ready that has the reducers converting the HalDoc to a plain object Campaign model instead of back in the services.

src/app/core/advertiser/advertiser.service.ts
sandikbarr: prettier can also be annoying, but it does keep the whitespace consistent

PR #141, opened 2020-03-02
adds flight form ngrx store

sandikbarr:
Refactors the flight state using NgRx. Removes a lot of the old RxJS store state and interaction. Almost a clean break.


package.json
```diff
@@ -33,7 +33,7 @@
     "core-js": "^2.5.4",
     "ngx-prx-styleguide": "2.20.0",
     "prx-ng-serve": "^1.1.0",
-    "rxjs": "~6.3.3",
+    "rxjs": "~6.5.4",
```
sandikbarr, line 36: updated to get dictionary support on forkJoin

src/app/campaign/campaign.component.ts
```diff
@@ -1,78 +1,76 @@
...
     public store: Store<any>,
     private accountService: AccountService,
     private advertiserService: AdvertiserService
   ) {}
 
   ngOnInit() {
-    this.campaignSaving$ = this.store.pipe(select(selectCampaignSaving));
     this.routeSub = this.route.paramMap
       .pipe(
+        tap(() => this.store.dispatch(new actions.CampaignLoadOptions())),
```
sandikbarr: This empty action starts the loading spinner. (Sometimes accounts take a long time to load and the campaign store isn't initialized/loaded until after the options lists because of that issue with reactive form select fields.)

src/app/campaign/campaign.component.ts
```diff
@@ -81,29 +79,20 @@ export class CampaignComponent implements OnInit, OnDestroy {
...
-              name: flightStates[id].localFlight.name || '(Flight)',
-              softDeleted: !!flightStates[id].softDeleted,
-              changed: flightStates[id].changed,
-              valid: flightStates[id].valid,
-              statusOk: !flightStates[id].localFlight.status || flightStates[id].localFlight.status === 'ok'
-            };
-          }
-        );
-      })
-    );
```
sandikbarr, line 106: The different states here are (see template above):
Loaded: Campaign and Flights have been new'ed or loaded via the api (show content or check error)
Loading: API is currently loading (show spinner)
Saving: Campaign and/or flights are saving. (show spinner on Save button)
Error: Campaign 404 or Flight not found on campaign (if loaded and error, show error)

src/app/campaign/campaign.component.ts
```diff
@@ -114,45 +103,15 @@ export class CampaignComponent implements OnInit, OnDestroy {
...
-
-    this.campaignStoreService.storeCampaign().subscribe(([changes, deletedDocs]) => {
-      this.toastr.success('Campaign saved');
-
-      // TODO: a better way to do this. like, much better.
-      // TODO: when on a flight with a new campaign, don't have the new campaign.id
-      //  so only route to flight if campaign already existed,
-      //  otherwise just let campaign.effects route to the new campaign for now
-      if (changes.id) {
-        const flightId = this.router.url.split('/flight/').pop();
```
sandikbarr, line 128: moved to campaign.effects!

src/app/campaign/flight/flight.container.ts
```diff
@@ -84,65 +82,55 @@ export class FlightContainerComponent implements OnInit, OnDestroy {
...
-        this.campaignStoreService.loadAvailability(state.flights[id].localFlight);
-        this.flightAvailability$ = this.campaignStoreService.getFlightAvailabilityRollup$(id);
-        this.flightDailyMin$ = this.campaignStoreService.getFlightDailyMin$(id);
-      }
-      this.currentFlightId = id;
-      this.flightState$.next(state.flights[id]);
-      this.currentInventoryUri$.next(state.flights[id].localFlight.set_inventory_uri);
-    } /*
-    else {
-      TODO: what to do about an invalid flight id and how to tell if a flight doesn't exist or hasn't yet been loaded into state
```
sandikbarr, line 99: _Fixed! forever!_ Once campaign with flights zoomed is loaded (or if total > count all flights also loaded), the error selector checks if the routed flight id exists on state

src/app/campaign/form/campaign-form.container.spec.ts
```diff
@@ -56,28 +59,23 @@ describe('CampaignFormContainerComponent', () => {
...
     const valid = false;
-    component.campaignUpdateFromForm({ campaign, changed, valid });
-    expect(campaignStoreService.setCampaign).toHaveBeenLastCalledWith({ localCampaign: campaign, changed, valid });
+    component.campaignUpdateFromForm({ campaign: campaignFixture, changed, valid });
+    expect(store.dispatch).toHaveBeenLastCalledWith(new actions.CampaignFormUpdate({ campaign: campaignFixture, changed, valid }));
   });
 
   it('sets the campaign after adding a new advertiser', () => {
     component.onAddAdvertiser('Squarespace');
-    expect(campaignStoreService.setCampaign).toHaveBeenCalled();
```
sandikbarr, line 81: whoops, missed this in the first campaign PR

src/app/campaign/store/effects/campaign.effects.ts
```diff
@@ -25,34 +25,124 @@ export class CampaignEffects {
...
+            createdFlightDocs: { [id: number]: HalDoc };
+          }) => new actions.CampaignSaveSuccess({ campaignDoc, deletedFlightDocs, updatedFlightDocs, createdFlightDocs })
+        ),
+        catchError(error => of(new actions.CampaignSaveFailure({ error })))
       );
     })
   );
 
-  constructor(private actions$: Actions, private campaignService: CampaignService, private router: Router) {}
+  @Effect()
```
sandikbarr, line 123: I used effects for add/dup flight because it generates the temporary id (nondeterministic) and routes to the new flight with the id.

src/app/campaign/store/effects/campaign.effects.ts
```diff
@@ -25,34 +25,124 @@ export class CampaignEffects {
...
-          //   // flight id changed (flight created)
-          //   this.router.navigate(['/campaign', campaign.id, 'flight', changes.flights[flightId]]);
-          // } else
-          if (payload.campaign.id !== campaignDoc.id) {
-            // campaign id changed (campaign created)
-            this.router.navigate(['/campaign', campaignDoc.id]);
-          }
+    ofType(ActionTypes.CAMPAIGN_SAVE),
+    map((action: actions.CampaignSave) => action.payload),
+    mergeMap((payload: CampaignFormSave) => {
```
sandikbarr, line 30: Save the campaign, then delete any softDeleted flights, update any changed flights, and create and new flights. Return results as HalDoc with flight ids mapped to their results.

src/app/campaign/store/effects/campaign.effects.ts
```diff
@@ -25,34 +25,124 @@ export class CampaignEffects {
...
+          }: {
+            campaignDoc: HalDoc;
+            deletedFlightDocs: { [id: number]: HalDoc };
+            updatedFlightDocs: { [id: number]: HalDoc };
+            createdFlightDocs: { [id: number]: HalDoc };
+          }) => {
+            this.toastr.success('Campaign saved');
+            // navigate to newly created campaign/flight or away from deleted flight
+            const campaignId = campaignDoc.id;
+            const flightId = +this.router.url.split('/flight/').pop();
```
sandikbarr, line 92: Could also have used the router store state to get the routed flight id or included it in the action, but you know what, this is fine (url.split.pop() taken from the component that was routing on save previously.)

src/app/campaign/store/selectors/combined.selectors.ts
```diff
@@ -0,0 +1,42 @@
...
+);
+
+export const selectFlightNotFoundError = createSelector(
+  selectCampaignLoaded,
+  selectRoutedFlightId,
+  selectFlightEntities,
+  (loaded, flightId, flights): string => loaded && flightId && !flights[flightId] && 'Flight Not Found'
+);
+
+export const selectError = createSelector(selectCampaignError, selectFlightNotFoundError, (campaignError: any, flightError: string) => {
```
sandikbarr, line 40: Could stand to do some better parsing off error messages, but we're displaying them now. (Make a change to a flight with a date in the past, get Error 422, see error shown in app)

src/app/campaign/store/selectors/flight.selectors.ts
```diff
@@ -0,0 +1,49 @@
...
+);
+export const selectRoutedFlightChanged = createSelector(
+  selectRoutedFlight,
+  (flightState: FlightState): boolean => flightState && flightState.changed
+);
+export const selectRoutedFlightDailyMinimum = createSelector(
+  selectRoutedFlight,
+  (flightState: FlightState): number => flightState && flightState.dailyMinimum
+);
+export const selectFlightDocById = createSelector(
```
sandikbarr, line 46: Today years old when I learned you could use props as the second argument to a selector to create a selector with parameters. This can be handy for when you don't have another selector/slice of state to combine (e.g. routed id) to get what you want from the store.

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -177,12 +128,11 @@ export class CampaignStoreService {
     );
   }
 
-  // the flightId parameter here will be the temp id in the case the flight has not yet been created
-  loadAvailability(flight: Flight, flightId?: string): Observable<Availability[]> {
+  // the flight.id parameter here will be the temp id in the case the flight has not yet been created
```
sandikbarr, line 134: Now uses flight.createdAt to determine if flight is new or already exists

src/app/core/user/user.service.ts
```diff
@@ -19,53 +19,52 @@ export class UserService {
...
-      })
-    ).subscribe(doc => {
-      this._userDoc.next(doc);
-    });
+    this.authService.token
+      .pipe(
+        first(),
+        concatMap(token => {
+          if (token) {
+            return combineLatest([of(token), this.userinfoService.getUserinfo()]);
```
sandikbarr, line 33: a bunch of prettier whitespace changes in this file along with this update to combineLatest parameters being an array for newer RxJS v6.5

src/app/campaign/campaign.component.ts
```diff
@@ -1,78 +1,75 @@
...
-    public store: Store<any>,
+    private store: Store<any>,
+    private campaignAction: CampaignActionService,
     private accountService: AccountService,
     private advertiserService: AdvertiserService
   ) {}
 
   ngOnInit() {
-    this.campaignSaving$ = this.store.pipe(select(selectCampaignSaving));
     this.routeSub = this.route.paramMap
```
sandikbarr, line 70: This is still a troublesome spot here that I do not love. From what I recall, we had issues with those select fields not populating with the loaded values if the options were not loaded before the form initialized?

Accounts and Advertisers could go into NgRx and be handled with campaign load on route change. Still some work to do to get all this cleaned up.

src/app/campaign/campaign.component.spec.ts
```diff
@@ -115,87 +88,30 @@ describe('CampaignComponent', () => {
...
   });
 
   it('submits the campaign forms', () => {
-    const campaign = campaignFactory();
-    campaignState.next(campaign);
+    jest.spyOn(campaignActionService, 'saveCampaignAndFlights');
     component.campaignSubmit();
-    expect(campaignStoreService.storeCampaign).toHaveBeenCalled();
-    expect(toastrService.success).toHaveBeenCalledWith('Campaign saved');
+    expect(campaignActionService.saveCampaignAndFlights).toHaveBeenCalled();
```
cavis, line 109: +1 checking that actions have been called really simplifies these tests.

src/app/campaign/campaign.component.ts
cavis: Maybe a good rainy day ticket?  (To move both into the NgRx).  Seems like we should be able to fix any load order dependencies then too.  Hopefully.  Maybe.

src/app/campaign/campaign.component.ts
cavis: +1

src/app/campaign/campaign.routing.ts
```diff
@@ -37,11 +39,14 @@ export const campaignComponents: any[] = [
...
   CampaignFormContainerComponent,
   CampaignFormComponent,
+  CampaignNavComponent,
   CampaignStatusComponent,
   FlightContainerComponent,
   FlightComponent,
   AvailabilityComponent,
-  GoalFormComponent
+  GoalFormComponent,
+  // to make the angular compiler happy, include all components that are defined in their own files, ugh fine
```
cavis, line 48: That's weird.  So the TestComponent will actually end up compiled into the prod js files then?

src/app/campaign/flight/flight.container.spec.ts
```diff
@@ -1,181 +1,157 @@
...
+import { flightFixture, flightDocFixture } from '../store/models/campaign-state.factory';
+import { reducers } from '../store';
+import * as actions from '../store/actions';
+import { CampaignActionService } from '../store/actions/campaign-action.service';
 
 import { FlightContainerComponent } from './flight.container';
+import { FlightComponent } from './flight.component';
+import { AvailabilityComponent } from '../availability/availability.component';
+import { GoalFormComponent } from '../availability/goal-form.component';
+import { TestComponent } from '../campaign-test.component';
```
cavis, line 33: SO MANY IMPORTS!!!  Guess that's what it takes to do non-stubbed testing here.  Thanks angular testing.

src/app/campaign/flight/flight.container.ts
```diff
@@ -38,44 +38,43 @@ import { map, filter, first, pluck } from 'rxjs/operators';
...
+    private campaignAction: CampaignActionService
   ) {}
 
   ngOnInit() {
+    this.flightLocal$ = this.store.pipe(select(selectRoutedLocalFlight));
+    this.softDeleted$ = this.store.pipe(select(selectRoutedFlightDeleted));
+    this.flightChanged$ = this.store.pipe(select(selectRoutedFlightChanged));
+    this.flightDailyMin$ = this.store.pipe(select(selectRoutedFlightDailyMinimum));
+    this.currentInventoryUri$ = this.store.pipe(select(selectCurrentInventoryUri));
+    this.flightAvailability$ = this.campaignStoreService.getFlightAvailabilityRollup$();
```
cavis, line 69: This is where availability is still coming from not-ngrx, right?

src/app/campaign/flight/flight.container.ts
cavis: Nice!  Forever!

src/app/campaign/nav/campaign-nav.component.ts
```diff
@@ -0,0 +1,36 @@
+import { Component, Input, Output, EventEmitter, ChangeDetectionStrategy } from '@angular/core';
+import { FlightState } from '../store/models';
+
+@Component({
+  selector: 'grove-campaign-nav',
```
cavis, line 5: Good call, moving this to its own component.

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  loadCampaign(id: number) {
+    this.store.dispatch(new actions.CampaignLoad({ id }));
+  }
+
+  updateCampaignForm(campaign: Campaign, changed: boolean, valid: boolean) {
+    this.store.dispatch(new actions.CampaignFormUpdate({ campaign, changed, valid }));
+  }
+
+  // tslint:disable-next-line: variable-name
```
cavis, line 69: At some point ... I fixed our HAL APIs to also accept the camelcase `setAdvertiserUri`.  I'll have to check if that ever made it into Augury.

src/app/campaign/store/effects/campaign.effects.spec.ts
```diff
@@ -113,23 +134,144 @@ describe('CampaignEffects', () => {
...
+      });
+      actions$.stream = hot('-a', { a: createAction });
+      const expected = cold('-r', { r: success });
+      expect(effects.campaignFormSave$).toBeObservable(expected);
+      expect(router.navigate).toHaveBeenCalledWith(['/campaign', campaignFixture.id, 'flight', flightFixture.id]);
+      done();
+    });
+  });
+
+  it('should redirect away from a deleted flight to the campaign', done => {
```
cavis, line 207: Nice!  Actually testing those edge cases!

src/app/campaign/store/effects/campaign.effects.ts
cavis: +1 encapsulating the insanity!

I do wonder how this handles errors.  I guess it would just short-circuit with the first non-okay response?  Maybe someday we'd want to run them all in parallel and let all succeed/fail before returning the error(s).

src/app/campaign/store/models/flight.models.ts
```diff
@@ -0,0 +1,35 @@
...
+  createdAt?: Date;
+  set_inventory_uri: string;
+}
+
+export interface FlightState {
+  id: number;
+  doc?: HalDoc;
+  localFlight: Flight;
+  remoteFlight?: Flight;
+  dailyMinimum?: number;
```
cavis, line 22: dailyMinimum is a `FlightState` attribute?  (not `Flight`)

src/app/campaign/store/selectors/combined.selectors.ts
cavis: There's also an Augury ticket out there to make the message part nicer - so may want to wait for that to land first.

src/app/campaign/store/selectors/flight.selectors.ts
cavis: Neat!

src/app/core/campaign/campaign.models.ts
```diff
@@ -1,31 +1,6 @@
 export interface CampaignState {
```
cavis, line 1: Nice to see these old models nearly-disappeared!

cavis (approved):
Left a couple questions - but this looks great to me!  Tests are especially cleaned up, with all the code moved out of the components.  Can't wait to run it locally, and watch those ngrx devtool store changes roll in!

src/app/campaign/campaign.routing.ts
sandikbarr: yes, this seems very dumb. probably needs to be in an excluded folder. not the last time I'll need a test component, will look into a better setup for this

src/app/campaign/flight/flight.container.spec.ts
sandikbarr: ugh, right =/ so much setup, so little to test

src/app/campaign/store/models/flight.models.ts
sandikbarr: For some reason we don't seem to save the dailyMinimum on the Flight. We only use it to submit to the availability preview. I thought that was odd too. Is it a property on the Flight and should it be saved there?

src/app/campaign/flight/flight.container.ts
sandikbarr: yep. The availability gets loaded when actions are called and the route changes (also in the action service), but the availability state still ends up in the ol' CampaignStoreService. Availability and allocations are next for the refactor though, so CampaignStoreService should disappear completely when this is done.

src/app/campaign/store/effects/campaign.effects.ts
sandikbarr: forkJoin will not return a result if any of the Observables has an error (or completes without emitting.) The error would bubble up into the stream and end up in this effect's catchError. The effects has to return an Observable of an action even in the event of error in order to keep the stream moving or the application itself would error with something like `sttream expected`. So we return a failure action. The error is saved onto the campaign state in teh campaign reducer, then it would be displayed through the error selector.

src/app/campaign/store/models/flight.models.ts
cavis: It is.  But I want to get rid of the field anyway, so i'm not too worried about it.

Issue #68, opened 2019-10-25
allocation preview validation errors


sandikbarr:
The allocation preview table should validate the total goal and daily minimum fields and give feedback to the user if either of them cannot be met. Row by row validation errors may turn out to be tricky in augury, so perhaps we just need to provide a useful error message.

sandikbarr:
HalRemote should pass error message from the backend if present.

sandikbarr:
On validation error, decode the JSON and display what went wrong. If it failed to allocate, re-run the allocation-preview to get the error displaying on the table. For other validation errors, display them somewhere global.

HAL handlers don't currently pass back the body when there's a response code > 300. These handlers are in Styleguide, so we need to either handle with a 200 response code, or make some changes in Styleguide to allow passing through the body of a non-200.

PR #156, opened 2020-03-23
allocation preview NgRx store

sandikbarr:
Sets up an NgRx store for allocation preview
 - The allocation preview is loaded whenever the flight start/end dates, series, zone, goal or daily minimum are changed
 - Allocation preview is loaded into an entity store keyed by `zone_date`
 - For now, the ol' campaign store is still providing the rollups that display allocation preview along with availability. The availability will be worked on next, and combining the two datasets with those rollups will be turned into NgRx selectors that provide the rollup data to the component.


src/app/campaign/store/effects/campaign.effects.spec.ts
```diff
@@ -87,14 +69,14 @@ describe('CampaignEffects', () => {
 
   it('should return campaign load failure action on error', () => {
```
sandikbarr, line 70: This was an annoying marble testing thing I had to work out with the error cases. I was just seeing a Complete in the expected stream instead of getting the Failure action. _But_ if I used an inner pipe with catchError in the effect instead of just piping through inline in the actions$, I did get the action in the test. I finally figured out that it was expecting a (grouping) of Failure with Complete `'(b|)'` for the Error `'#'`.

Fixed.

src/app/campaign/store/models/allocation-preview.models.ts
```diff
@@ -0,0 +1,28 @@
+import { HalDoc } from 'ngx-prx-styleguide';
+
+export interface AllocationPreview {
+  date: Date;
+  goalCount: number;
+  inventoryDayId: number;
+  zone: string;
+}
+
```
sandikbarr, line 9: The Allocation Preview response is a bit of an odd 🦆 in our APIs where you might expect it to have items embedded for the allocation days, but what we actually have in the response is an object with (_most_) of the params given in the request and an allocations array that contains the individual allocations days. When translated to an entity reducer model, the allocation days are the entities and the params are held on the root of the store's state.

[Re: _most_, the set_inventory_uri isn't returned in the response, and where we're using that temporary flight ids for new flights, that won't be in the response (or the request) either. So these two properties are saved on the state with the LOAD action.]

src/app/campaign/availability/availability.component.ts
```diff
@@ -96,8 +96,8 @@ export class AvailabilityComponent {
 
     // Check for allocation preview error.
     // TODO: Updated with discussed "nice_message" when available.
-    if (this.allocationPreviewError) {
-      errors.push(`Got error ${this.allocationPreviewError.status} from allocation preview.`);
```
sandikbarr, line 100: It's now showing any allocation preview error messages attached to the error response body. I am going to close #68 with this PR, and we can worry about formatting later.

src/app/campaign/campaign.component.spec.ts
```diff
@@ -20,7 +20,7 @@ import { CampaignActionService } from './store/actions/campaign-action.service';
 import { CampaignComponent } from './campaign.component';
 import { CampaignStatusComponent } from './status/campaign-status.component';
 import { CampaignNavComponent } from './nav/campaign-nav.component';
-import { TestComponent, campaignRoutes } from './campaign-test.component';
+import { TestComponent, campaignRoutes } from '../../testing/test.component';
```
sandikbarr, line 24: Got that TestComponent put in a place where it will be ignored in the angular build.

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -50,12 +50,9 @@ export class CampaignStoreService {
               (acc, day) => {
                 // get the day's allocationPreview from the allocationPreview state
                 day.allocationPreview =
-                  state.allocationPreview &&
-                  state.allocationPreview[`${flightId}`] &&
-                  state.allocationPreview[`${flightId}`][availability.zone] &&
-                  state.allocationPreview[`${flightId}`][availability.zone].allocations &&
-                  state.allocationPreview[`${flightId}`][availability.zone].allocations[day.startDate] &&
-                  state.allocationPreview[`${flightId}`][availability.zone].allocations[day.startDate].goalCount;
```
sandikbarr, line 58: for now we just grab the allocation preview from the store's entities here instead

src/app/core/campaign/campaign-store.service.ts
```diff
@@ -77,10 +74,12 @@ export class CampaignStoreService {
                   // accumulate values onto week
                   const weekTotals = acc.totals.groups[weekBeginString];
                   acc.totals.groups[weekBeginString] = {
+                    // allocated and availability will be null rather than undefined, null + 0 === 0
                     allocated: weekTotals.allocated ? weekTotals.allocated + day.allocated : day.allocated,
                     availability: weekTotals.availability ? weekTotals.availability + day.availability : day.availability,
                     allocationPreview: weekTotals.allocationPreview
-                      ? weekTotals.allocationPreview + day.allocationPreview
```
sandikbarr, line 83: addressed a bit of a bug here where we we're getting NaNs with summing undefined values

src/app/campaign/availability/availability.component.ts
cavis: +1

src/app/campaign/campaign.component.spec.ts
cavis: Nice!

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -45,41 +46,78 @@ export class CampaignActionService implements OnDestroy {
...
+    createdAt: Date,
+    // tslint:disable-next-line: variable-name
+    set_inventory_uri: string,
+    name: string,
+    startAt: Date,
+    endAt: Date,
+    totalGoal: number,
+    dailyMinimum: number,
+    zones: string[]
+  ) {
```
cavis: Why a bunch of individual params, vs an object for this function call?

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -45,41 +46,78 @@ export class CampaignActionService implements OnDestroy {
     if (dateRangeValid && availabilityParamsChanged) {
       this.campaignStoreService.loadAvailability(formFlight);
       if (formFlight.totalGoal) {
-        this.campaignStoreService.loadAllocationPreview(formFlight, dailyMinimum);
+        this.loadAllocationPreview(
```
cavis: Yeah, this feels like a lot of params to be passing in an exact order.

src/app/campaign/store/effects/campaign.effects.spec.ts
cavis: /me head explodes.  I guess I should do some more reading/tutorials on test marbles!

src/app/core/campaign/campaign-store.service.ts
sandikbarr: This code is fragile and will be adjusted in the refactor. I just pushed up another commit to address this bug where the allocation preview thinks it should be showing but there aren't value (because false >= 0 is what the component was comparing.)

src/app/campaign/store/effects/campaign.effects.ts
```diff
@@ -14,19 +14,16 @@ export class CampaignEffects {
...
     ofType(ActionTypes.CAMPAIGN_LOAD),
-    map((action: actions.CampaignLoad) => action.payload),
-    mergeMap(payload =>
-      this.campaignService.loadCampaignZoomFlights(payload.id).pipe(
-        map(({ campaignDoc, flightDocs }) => new actions.CampaignLoadSuccess({ campaignDoc, flightDocs })),
-        catchError(error => of(new actions.CampaignLoadFailure({ error })))
-      )
-    )
+    map((action: campaignActions.CampaignLoad) => action.payload),
+    mergeMap(payload => this.campaignService.loadCampaignZoomFlights(payload.id)),
```
cavis, line 18: This is functionally equivalent to the above, right?  (With the 2nd pipe inside the mergemap)

src/app/campaign/store/models/allocation-preview.models.ts
cavis: Makes sense.  At some point, we decided `flight.allocations` are literally attributes of the flight - not a paged collection related to the flight.  Is a bit strange ... but it beats trying to embed/page through them!

It'll be the same thing with `flight.zones` and `flight.targets` ... we're abstracting over the relation and calling them attributes on the flight.  So the API for setting them is a bit more sane.

cavis (approved):
+12.  Looks much better!  I had one comment/suggestion about the number of params on `loadAllocationPreview()` (seems like it should just take in an object at this point?).  But that's it!

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: I had the same thought, and then I thought do I really need this extra wrapper function anyway. It makes testing a bit more straightforward to have these wrapper functions rather than calling store.dispatch directly, but I might still get rid of the ones that are just wrappers around dispatch + action creator. I'm still having thoughts on this part. It's not a standard for NgRx and feels a bit extra. The service is necessary, or something that can run and watch those state changes for loading data, and better here than in components, but the rest feels like extra boilerplate.

src/app/campaign/store/effects/campaign.effects.ts
sandikbarr: yep, same functionality. Best practice says we should use a single pipe and do one operation after the next.

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: @cavis addressed this one by using the payload interface provided by the action creator and removing some other wrappers around store.dispatch that just felt extra

cavis (approved):
👍

PR #157, opened 2020-03-24
fixes sort flights by createdAt after form update

sandikbarr:
The flight tabs are sorted by the localFlight.createdAt property. The createAt property doesnt exist in the form so it was getting cleared out of the localFlight on update (could've swore I looked at this earlier and thought it was good, 🤷‍♀ fixed now).


cavis (approved):
👍 this could be helpful in other instances too (if we introduce a writeable non-null param into the Augury API ... but don't tell Grove about it right away).

Issue #94, opened 2019-12-16
MP3 url field


cavis:
After PRX/augury.prx.org#148 ... add the "creative url" field to the flight form.  (Next to the zone).

No uploader for now.  Just paste in a URL.  Validate it's a real-url.  For this epic, only allow _one_ zone_name and _one_ mp3 url.  We'll tackle multi-zone/companions later.

_MAYBE_ validate it's an mp3.

![image](https://user-images.githubusercontent.com/1410587/70939919-7e5ae480-2006-11ea-9147-b3c78f43c321.png)

cavis:
Noting: the "flight zones" API is going to change slightly, breaking the frontend.  Will attempt to merge these at the same time.

PR #161, opened 2020-04-01
Flight zone creatives

cavis:
Updated PR for #94.  Must be deployed concurrently with PRX/augury.prx.org#203.

I put a [demo video here](http://cav.is/OFBLK6t.mp4)

image

Basically, the Flights API went from `zones: ['pre_1', 'pre_2']` to: 

```
{
  zones: [
    {id: 'pre_1', label: 'Preroll 1'],
    {id: 'pre_2', label: 'Preroll 2', url: 'http://some/file.mp3'}
  ]
}
```

The only exception is the flight-preview endpoint, which still takes in the old `?zones=pre1,pre2` query params.  I've noted we should probably update that over in PRX/augury.prx.org#206.


src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -38,6 +38,10 @@ export class CampaignActionService implements OnDestroy {
       });
   }
 
+  private getZoneIds(zones: any[]): string[] {
+    return zones.map(z => (z ? z.id || z : z)).filter(z => z);
```
sandikbarr: why not filter first and avoid the ternary?

src/app/core/allocation/allocation-preview.service.ts
```diff
@@ -32,9 +32,13 @@ export class AllocationPreviewService {
...
         totalGoal,
         dailyMinimum,
-        zones
+        zones: this.getZoneIds(zones)
       });
     }
     return preview;
   }
+
+  private getZoneIds(zones: any[]): string[] {
```
sandikbarr: maybe make this a static function in the flight.model (and filter before the map)

src/app/shared/pipes/zone-label.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({ name: 'zoneLabel' })
+export class ZoneLabelPipe implements PipeTransform {
+  transform(value: any) {
+    if (Array.isArray(value)) {
+      return value.map((v: any) => this.transform(v)).join(', ');
+    } else if (value) {
+      return value.label || value.name || value.zoneName || value.id || value;
```
sandikbarr, line 9: fallbacks on fallbacks! but this puts the mess in one place 😬👍

src/app/campaign/flight/flight.component.ts
```diff
@@ -86,13 +73,27 @@ export class FlightComponent implements OnInit {
     });
   }
 
+  zoneOptionsFiltered(zoneIndex?: number) {
+    const myZones = (this.flight && this.flight.zones) || [];
+    const options = this.zoneOptions || <InventoryZone[]>myZones;
```
sandikbarr, line 82: maybe add a comment to this one explaining why it's either keeping or filtering out the zone from the options relevant to the zone index in the flight

sandikbarr (approved):
couple minors nits and I'm just not seeing the reactive form issue without the boolean guard but it's not like that's not been an issue, and this works!

sandikbarr (approved):

PR #164, opened 2020-04-03
Update readme with definitions for commonly used terms

brandonhundt:
@cavis and I, (mostly Ryan) have provided definitions for commonly used terms within the app. We think these should live in Grove's readme until we have more standard docs available.


sandikbarr (approved):
👍!

Issue #76, opened 2019-10-31
Advertiser validation


sandikbarr:
If a user has typed a new advertiser entry into the autocomplete, the form validation should not allow Saving the campaign and flights if the advertiser has not yet been added.

PR #166, opened 2020-04-06
Fixes loading dependencies

sandikbarr:
* adds stores for advertisers and accounts
* sets up dependency by not showing campaign form until campaign, advertisers, and accounts are in state
* fixes user.service bug and toJSON() of null error
* fixes new advertiser validation: #76


src/app/campaign/form/campaign-form.container.ts
```diff
@@ -1,32 +1,36 @@
...
 import * as campaignActions from '../store/actions/campaign-action.creator';
+import * as advertiserActions from '../store/actions/advertiser-action.creator';
 
 @Component({
   selector: 'grove-campaign-form.container',
   template: `
     <grove-campaign-form
-      [campaign]="campaign$ | async"
-      [accounts]="accounts$ | async"
-      [advertisers]="advertisers$ | async"
```
sandikbarr, line 15: don't show the form until dependent data is loaded:

src/app/core/user/user.service.ts
```diff
@@ -14,7 +14,7 @@ export class UserService {
   userinfo: Userinfo;
 
   // tslint:disable-next-line: variable-name
-  private _userDoc = new BehaviorSubject(null);
```
sandikbarr, line 17: The toJSON() of null error, this is the real bug fix here: I switched from BehaviorSubject to ReplaySubject because the _userDoc should not have an initial value. The userDoc Observable shouldn't emit until the user has been loaded.

src/app/campaign/form/campaign-form.component.ts
```diff
@@ -121,4 +121,12 @@ export class CampaignFormComponent implements OnInit {
   onAddAdvertiser(name: string) {
     this.addAdvertiser.emit(name);
   }
+
```
sandikbarr, line 124: validate that a new (user entered value not in list) advertiser has been added:

src/app/campaign/form/campaign-form.container.ts
```diff
@@ -1,32 +1,36 @@
...
+      *ngIf="{ campaigns: campaign$ | async, accounts: accounts$ | async, advertisers: advertisers$ | async } as data"
```
cavis, line 14: Interesting pattern - have not seen this before!  Makes sense though.  And nice to encapsulate it to the container (and out of the form).

src/app/campaign/form/campaign-form.container.ts
```diff
@@ -35,12 +39,6 @@ export class CampaignFormContainerComponent implements OnInit {
   }
 
   onAddAdvertiser(name: string) {
-    const post = this.advertiserService.addAdvertiser(name);
-
-    post.subscribe(result => {
-      this.store.dispatch(new campaignActions.CampaignSetAdvertiser({ set_advertiser_uri: result.set_advertiser_uri }));
-      // TODO: how are we notifying in this app, material or the ol' toast?
-      // this.toastr.success('Advertiser added');
```
cavis, line 43: I'm just curious ... if we did want to toastr on advertiser create, where would you wire that in?  (Some sort of advertiser-create-success-selector?)

src/app/campaign/store/models/campaign-state.factory.ts
```diff
@@ -80,7 +128,7 @@ export const allocationPreviewParamsFixture = {
   endAt: flightFixture.endAt,
   name: flightFixture.name,
   set_inventory_uri: flightFixture.set_inventory_uri,
-  zones: flightFixture.zones,
+  zones: flightFixture.zones.map(zone => ({ id: zone.id })),
```
cavis, line 131: Removing the labels from these?  Or are there other properties messing things up?  (Does seem like per PRX/augury.prx.org#206 this can someday just be the Flight model).

src/app/campaign/store/reducers/campaign.reducer.ts
```diff
@@ -39,17 +38,19 @@ export function reducer(state = initialState, action: CampaignActions): Campaign
...
+      const { set_advertiser_uri } = docToAdvertiser(action.payload.doc);
       return {
         ...state,
         localCampaign: {
           ...state.localCampaign,
           set_advertiser_uri
         },
-        changed: true
+        changed: true,
+        valid: true
```
cavis: Hmm... can you really assume the form is valid here?  There could be other missing+required fields.

cavis (changes requested):
One comment about marking the form valid after advertiser-add.  Seems that might introduce a bug?  Otherwise :+82:

(Or maybe there's no way around that, since validations come out of the form).

src/app/campaign/form/campaign-form.container.ts
sandikbarr: Effects! I've got a toast in there for Campaign Save when I knew I had a successful PUT/POST with the API.

src/app/campaign/store/models/campaign-state.factory.ts
sandikbarr: Yeah... tbh I do not love what happened to the allocation preview params zone with that recent zone mp3 url stuff. You changed that zone param to an object with just the id, so I had to make some adjustments in a few places, was thinking about changing it back to just the id string. https://github.com/PRX/grove.prx.org/blob/master/src/app/campaign/store/reducers/allocation-preview.reducer.ts#L15

src/app/campaign/store/reducers/campaign.reducer.ts
sandikbarr: valid concern! it should just keep valid from existing state. will fix.

src/app/campaign/store/models/campaign-state.factory.ts
cavis: Yeah, I was wiring things through as if 206 was already done, until I realized that I couldn't just serialize the entire Flight model as-is.  (And never undid that change).  I'm fine with changing back to array-of-strings until 206 hits.

src/app/campaign/form/campaign-form.container.ts
cavis: Oh right!  So you can directly trigger the toast without a component.  Yeah, that sounds cleaner than what I was thinking.

I always forget about effects!

src/app/campaign/store/models/campaign-state.factory.ts
sandikbarr: ah ok if that preview endpoint is gonna take all those params then let's leave an an object and we'll sort it out later

src/app/campaign/form/campaign-form.container.ts
sandikbarr: I think that's how those side effects are kinda supposed to work, once you got em handled you don't have to think about em. ;)

cavis (approved):

PR #171, opened 2020-04-14
duplicate campaign

### LINKED ISSUE #42: duplicate campaign
From Brandon:

> Ad/ops often runs the same campaign with the same flight each month, but only changes the dates. You can initiate a duplicated campaign from the dropdown arrow next to the save button on the inventory form. There is also an option to duplicate a campaign from the campaign dashboard view.

sandikbarr:
Closes #42 

Duplicates a campaign either from the campaign form or from the dashboard. Both methods are implemented as routerLinks to `/campaign/new` with state. Campaign/flights from the form can be duplicated synchronously, but campaign/flights from the dashboard are duplicated by id with an asynchronous effect to create the success action that populates state.

Also fixed, a nasty bug in the campaign.service that is currently preventing new flights on new campaigns to be persisted. I removed the campaign HalDoc observable from the campaign.service and am providing it as a parameter to those functions that make Hal API calls. For campaigns that had not yet been persisted, this Observable didn't fire until it saw a campaign HalDoc on the state and then the rest of the Observable chain to create those new flights would fire off. Ugly.

### DIFF
--- a/src/app/campaign/campaign.component.ts
+++ b/src/app/campaign/campaign.component.ts
@@ -2,9 +2,10 @@ import { Component, ChangeDetectionStrategy, OnInit, OnDestroy } from '@angular/
 import { ActivatedRoute } from '@angular/router';
 import { Observable, Subscription } from 'rxjs';
 import { Store, select } from '@ngrx/store';
-import { FlightState } from './store/models';
+import { FlightState, Campaign } from './store/models';
 import {
   selectError,
+  selectLocalCampaign,
   selectCampaignLoaded,
   selectCampaignLoading,
   selectCampaignSaving,
@@ -30,7 +31,14 @@ import { CampaignActionService } from './store/actions/campaign-action.service';
     ></grove-campaign-status>
     <mat-drawer-container autosize>
       <mat-drawer role="navigation" mode="side" opened disableClose>
-        <grove-campaign-nav [flights]="flights$ | async" (createFlight)="createFlight()"></grove-campaign-nav>
+        <grove-campaign-nav
+          [campaign]="campaign$ | async"
+          [flights]="flights$ | async"
+          [valid]="valid$ | async"
+          [changed]="changed$ | async"
+          [isSaving]="campaignSaving$ | async"
+          (createFlight)="createFlight()"
+        ></grove-campaign-nav>
       </mat-drawer>
       <mat-drawer-content role="main">
         <ng-container *ngIf="campaignLoaded$ | async; else loading">
@@ -49,6 +57,7 @@ import { CampaignActionService } from './store/actions/campaign-action.service';
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class CampaignComponent implements OnInit, OnDestroy {
+  campaign$: Observable<Campaign>;
   flights$: Observable<FlightState[]>;
   campaignLoaded$: Observable<boolean>;
   campaignLoading$: Observable<boolean>;
@@ -69,9 +78,17 @@ export class CampaignComponent implements OnInit, OnDestroy {
         const id = +params.get('id');
         this.store.dispatch(new campaignActions.CampaignLoad({ id }));
       } else {
-        this.store.dispatch(new campaignActions.CampaignNew());
+        const { state } = window.history;
+        if (state.campaign) {
+          this.store.dispatch(new campaignActions.CampaignDupFromForm({ campaign: state.campaign, flights: state.flights }));
+        } else if (state.id) {
+          this.store.dispatch(new campaignActions.CampaignDupById({ id: state.id }));
+        } else {
+          this.store.dispatch(new campaignActions.CampaignNew({}));
+        }
       }
     });
+    this.campaign$ = this.store.pipe(select(selectLocalCampaign));
     this.campaignLoaded$ = this.store.pipe(select(selectCampaignLoaded));
     this.campaignLoading$ = this.store.pipe(select(selectCampaignLoading));
     this.campaignSaving$ = this.store.pipe(select(selectCampaignSaving));
--- a/src/app/campaign/flight/flight.component.ts
+++ b/src/app/campaign/flight/flight.component.ts
@@ -69,7 +69,7 @@ export class FlightComponent implements OnInit {
   ngOnInit() {
     this.flightForm.valueChanges.subscribe(cmp => {
       // preserving id on the flight because doesn't exist in form fields
-      this.formStatusChanged({ ...cmp, id: this.flight.id });
+      this.formStatusChanged({ ...cmp, id: this.flight.id }, this.flightForm.dirty);
     });
   }
 
@@ -79,7 +79,7 @@ export class FlightComponent implements OnInit {
     const myZones = (this.flight && this.flight.zones) || [];
 
     // if zoneOptions hasn't loaded yet, use flight.zones as options
-    const options = this.zoneOptions || <InventoryZone[]>myZones;
+    const options = this.zoneOptions || (myZones as InventoryZone[]);
     return options.filter(zone => {
       if (myZones[zoneIndex] && myZones[zoneIndex].id === zone.id) {
         return true;
@@ -90,29 +90,34 @@ export class FlightComponent implements OnInit {
   }
 
   // emits updates when reactive form fields change
-  formStatusChanged(flight?: Flight) {
+  formStatusChanged(flight: Flight, changed?: boolean) {
     if (this.emitFormUpdates) {
       this.flightUpdate.emit({
         flight: { ...flight, totalGoal: this.flight.totalGoal },
-        changed: this.flightForm.dirty,
+        changed,
         valid: this.flightForm.valid
       });
     }
   }
 
   onDateRangeChange({ startAt, endAt }: { startAt?: Date; endAt?: Date }) {
-    if (this.emitFormUpdates) {
-      this.flightUpdate.emit({
-        flight: {
-          ...this.flight,
-          ...(startAt && { startAt }),
-          ...(endAt && { endAt }),
-          totalGoal: this.flight.totalGoal
-        },
-        changed: true,
-        valid: this.flightForm.valid
-      });
+    // these dates pickers are not actually connected to the reactive form
+    // so in order for the flightForm to be valid on emit, these values must be set on the form
+    if (startAt) {
+      this.flightForm.get('startAt').setValue(startAt, { emitEvent: false });
     }
+    if (endAt) {
+      this.flightForm.get('endAt').setValue(endAt, { emitEvent: false });
+    }
+    this.formStatusChanged(
+      {
+        ...this.flight,
+        ...(startAt && { startAt }),
+        ...(endAt && { endAt }),
+        totalGoal: this.flight.totalGoal
+      },
+      true
+    );
   }
 
   // updates the form from @Input() set flight
--- a/src/app/campaign/nav/campaign-nav.component.ts
+++ b/src/app/campaign/nav/campaign-nav.component.ts
@@ -22,7 +22,12 @@ import { Campaign, Flight, FlightState } from '../store/models';
       </a>
     </mat-nav-list>
     <a class="secondary-link" routerLink="" (click)="createFlight.emit()"><mat-icon>add</mat-icon> Add a Flight</a>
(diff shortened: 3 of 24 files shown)


src/app/core/campaign/campaign.service.ts
```diff
@@ -41,16 +33,16 @@ export class CampaignService {
...
+    return doc.update(campaign);
   }
 
   createCampaign(campaign: Campaign): Observable<HalDoc> {
-    return this.augury.root.pipe(switchMap(rootDoc => rootDoc.create('prx:campaign', {}, campaign)));
+    return this.augury.root.pipe(mergeMap(rootDoc => rootDoc.create('prx:campaign', {}, campaign)));
   }
 
-  createFlight(flight: Flight): Observable<HalDoc> {
-    return this.campaignDoc$.pipe(switchMap(doc => doc.create('prx:flights', {}, flight)));
```
sandikbarr, line 53: Providing the campaign HalDoc as a parameter instead of hooking up to the state Observable fixes a nasty bug where new flights on new campaigns currently cannot be saved. The HalDoc observable wasn't firing until it saw a value from the persisted campaign, so the flight POST events weren't happening until some time later if you loaded a different campaign and they were posting to the wrong campaign. Gross. Didnt like that state tied up in here anyway.

src/app/campaign/campaign.component.ts
```diff
@@ -69,9 +78,17 @@ export class CampaignComponent implements OnInit, OnDestroy {
         const id = +params.get('id');
         this.store.dispatch(new campaignActions.CampaignLoad({ id }));
       } else {
-        this.store.dispatch(new campaignActions.CampaignNew());
```
sandikbarr, line 72: Both duplicated and new campaigns are routed to `/campaign/new`. Angular 7.2 supports a `routerLink` state we can use to pass data between routes: https://netbasal.com/set-state-object-when-navigating-in-angular-7-2-b87c5b977bb.

src/app/campaign/flight/flight.component.ts
```diff
@@ -90,29 +90,34 @@ export class FlightComponent implements OnInit {
...
-      this.flightUpdate.emit({
-        flight: {
-          ...this.flight,
-          ...(startAt && { startAt }),
-          ...(endAt && { endAt }),
-          totalGoal: this.flight.totalGoal
-        },
-        changed: true,
-        valid: this.flightForm.valid
-      });
```
sandikbarr, line 114: I was having problems with invalid form state on these dates when they were empty for duplicated campaigns. I saw they weren't yet set at this point in the reactive form because these datapickers are not reactive form connected. I thought it should call the same formStatusChanged function. We have several tickets and some open PRs that I need to get back to at some point on the material date range pickers and reactive form start and end date validation.

src/app/campaign/nav/campaign-nav.component.ts
```diff
@@ -21,16 +21,33 @@ import { FlightState } from '../store/models';
         <mat-icon color="warn" *ngIf="!statusOk(flight)">priority_high</mat-icon>
       </a>
     </mat-nav-list>
-    <a class="add-flight" [routerLink]="" (click)="createFlight.emit()"><mat-icon>add</mat-icon> Add a Flight</a>
+    <a class="secondary-link" routerLink="" (click)="createFlight.emit()"><mat-icon>add</mat-icon> Add a Flight</a>
```
sandikbarr, line 24: maybe at some point we'll decide to "Save & Duplicate" campaigns in one transaction(?) if they've made changes or not yet persisted the campaign, but at this point it's not allowed

src/app/campaign/store/actions/campaign-action.creator.ts
```diff
@@ -5,6 +5,36 @@ import { HalDoc } from 'ngx-prx-styleguide';
...
 export class CampaignNew implements Action {
   readonly type = ActionTypes.CAMPAIGN_NEW;
+
+  constructor(public payload: {}) {}
+}
+
+export class CampaignDupFromForm implements Action {
+  readonly type = ActionTypes.CAMPAIGN_DUP_FROM_FORM;
+
+  constructor(public payload: { campaign: Campaign; flights: Flight[]; timestamp?: number }) {
```
sandikbarr, line 15: Following redux principles, the timestamp (nondeterministic) is created outside the reducer and the reducer uses it to generate the temporary flight ids keeping the reducer pure.

src/app/campaign/store/effects/campaign.effects.ts
```diff
@@ -61,7 +61,7 @@ export class CampaignEffects {
             deletedFlightDocs: { [id: number]: HalDoc };
           }) => {
             const createdFlights: { [id: number]: Observable<HalDoc> } = payload.createdFlights.reduce(
-              (acc, flight) => ({ ...acc, [flight.id]: this.campaignService.createFlight(flight) }),
```
sandikbarr, line 64: Passing in the campaignDoc now to fix this nasty bug: https://github.com/PRX/grove.prx.org/pull/171/files#r408448986

src/app/campaign/campaign.component.ts
cavis: Ooh, that's cool.

src/app/campaign/flight/flight.component.ts
```diff
@@ -90,29 +90,34 @@ export class FlightComponent implements OnInit {
...
+    }
+    if (endAt) {
+      this.flightForm.get('endAt').setValue(endAt, { emitEvent: false });
     }
+    this.formStatusChanged(
+      {
+        ...this.flight,
+        ...(startAt && { startAt }),
+        ...(endAt && { endAt }),
+        totalGoal: this.flight.totalGoal
```
cavis, line 117: Still a bit weird that this field is part of a different form component.

src/app/campaign/flight/flight.component.ts
cavis: +1.  Would definitely simplify things to make these reactive form date fields.

src/app/campaign/store/actions/campaign-action.creator.ts
cavis: Makes sense.  I'm guessing then the real code never hits the `|| Date.now()` below.  (Maybe the tests do though).

src/app/campaign/store/reducers/campaign.reducer.ts
```diff
@@ -26,6 +26,26 @@ export function reducer(state = initialState, action: CampaignActions | Advertis
     case ActionTypes.CAMPAIGN_NEW: {
       return { ...initialState, loading: false, loaded: true };
     }
+    case ActionTypes.CAMPAIGN_DUP_FROM_FORM: {
+      return {
+        localCampaign: duplicateCampaign(action.payload.campaign),
+        changed: true,
+        valid: true,
```
cavis, line 33: These are valid?  I guess if you were only allowing already-valid campaigns to be duplicated.

src/app/core/campaign/campaign.service.ts
cavis: Yuck - good catch!

src/app/campaign/store/models/campaign.models.ts
```diff
@@ -32,3 +32,10 @@ export const docToCampaign = (doc: HalDoc): Campaign => {
   campaign.set_account_uri = doc.expand('prx:account');
   return campaign;
 };
+
+export const duplicateCampaign = (campaign: Campaign): Campaign => {
+  // remove id from dup campaign and set to Draft
+  const { id, ...dupCampaign } = campaign;
```
cavis, line 38: Maybe more of a UX / @brandonhundt thing ... but I would prefer if the duplicated flights/campaigns added `"DUPLICATE #{prev_title}"` to the title or something.  Otherwise a bit hard to tell what happened - this thing works so fast, and the page hardly changes, so it's hard to tell what just happened when I click duplicate.

cavis (approved):
:+12: Couple of questions, but this looks/works good to me.  Some nice improvements to how flight-duplicating is handled in the code as well!

src/app/campaign/flight/flight.component.ts
sandikbarr: I imagine to consolidate these into a single form the form needs to live in the form.container with these form group components inside it. I wish Angular made that easier. Also, I don't think I'll be allowed to make anymore big refactoring changes like that at this time.

src/app/campaign/store/actions/campaign-action.creator.ts
sandikbarr: 😬 so it's actually the opposite, the tests provide a timestamp so they can validate against it for the flightIds, and the application expects the action to generate it. With how I've been doing things, I probably would have just provided it as part of the original payload object in the application, but I didn't because tbh I made this decision about where to actually put this code last. And I do kind of like this. I had forgotten about the idea of using action creators to generate this type of nondeterministic data. I've used it before, and it just kind of clicked again when working on this. So yeah, this is introducing a new idea to the application, but from my recollection, it is a standard practice.

cavis: Ohh... okay.  Yeah, I follow.  I do that kind of thing all the time for testing (inject the timestamp, but default to Time.now()).

src/app/campaign/store/models/campaign.models.ts
sandikbarr: This is good thinking. Duplicated Flights (within the campaign) do get the suffix of " (Copy)" appended to their name. And/or maybe it should have a confirmation, "Are you sure you want to duplicate?" Or a toast notification that the duplication has been successful. It almost shouldnt be a disappearing toast though because I feel like it should contain instruction to set the flight dates. I don't feel like those interactions haven't been thought about too much in this application.

src/app/campaign/store/actions/campaign-action.creator.ts
sandikbarr: I am really not proud of this but one day I found myself in a spot and figured out how to mock Date.now in specs https://github.com/PRX/grove.prx.org/blob/master/src/app/campaign/store/effects/campaign.effects.spec.ts#L258

src/app/campaign/store/reducers/campaign.reducer.ts
sandikbarr: It's only valid campaigns that can be duplicated. @brandonhundt and I talked this one out https://prx.slack.com/archives/GCEJE5136/p1586446627005200?thread_ts=1586444042.004000&cid=GCEJE5136 

Also I dont have any validations written to validate these models except what we get from the reactive form, so I have to assume either valid or invalid or duplicate all the reactive form validations. Maybe we will need to have separate validations against the model at some point, but I dont think this warrants it just yet.

src/app/campaign/store/actions/campaign-action.creator.ts
cavis: Ha!  Mocking Date.now is like a drug.  Now you won't be able to stop.

src/app/campaign/flight/flight.component.ts
cavis: Makes sense.  This isn't terrible, considering a gigantic single form component is the alternative.  Just a bit strange.

Maybe a potential refactor when we start adding other fields to that sidebar!

src/app/campaign/store/reducers/campaign.reducer.ts
cavis: +1 makes sense.

src/app/campaign/store/models/campaign.models.ts
cavis: Just my thoughts!  I'm fine with merging this and figuring it out later too.

sandikbarr: Honestly, yeah I think we'll have a few afterthoughts like that on this project. I haven't been feeling like things are fully flushed out UXwise.

src/app/campaign/flight/flight.component.ts
sandikbarr: _maybe_ but I spent my refactoring capital on NgRx, so it probably won't happen in the near term.

sandikbarr:
Added a toast notification onto this. Will merge when the build finishes.

cavis:
👍

PR #172, opened 2020-04-16
flight form id and other fixes

sandikbarr:
* Fixes the flight form id and removes the boolean valueChanges guard. I fixed a couple other things in testing, and I feel confident about removing that guard.
* Cleans up the FlightState and Flight models to use the localFlight.id instead of having that id field duplicated into FlightState
* Sets the dailyMinimum on flights instead of just using it for preview
* Also sorts advertisers by name to improve the autocomplete selection. I couldn't replicate the bug with adding new advertisers instead of selecting existing ones, but I did find selection awkward with similarly named advertisers when they weren't sorted.

### DIFF
--- a/src/app/campaign/availability/goal-form.component.ts
+++ b/src/app/campaign/availability/goal-form.component.ts
@@ -31,20 +31,10 @@ export class GoalFormComponent implements OnInit, OnDestroy {
   set flight(flight: Flight) {
     if (flight) {
       this._flight = flight;
-      this.updateGoalForm({ totalGoal: this._flight.totalGoal });
+      this.updateGoalForm({ totalGoal: this._flight.totalGoal, dailyMinimum: this._flight.dailyMinimum });
     }
   }
-  // tslint:disable-next-line
-  private _dailyMinimum: number;
-  get dailyMinimum(): number {
-    return this._dailyMinimum;
-  }
-  @Input()
-  set dailyMinimum(dailyMinimum: number) {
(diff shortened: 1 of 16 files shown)


src/app/campaign/availability/goal-form.component.ts
```diff
@@ -31,20 +31,10 @@ export class GoalFormComponent implements OnInit, OnDestroy {
...
-  get dailyMinimum(): number {
-    return this._dailyMinimum;
-  }
-  @Input()
-  set dailyMinimum(dailyMinimum: number) {
-    this._dailyMinimum = dailyMinimum;
-    this.updateGoalForm({ dailyMinimum });
-  }
-  @Output() goalChange = new EventEmitter<{ flight: Flight; dailyMinimum: number }>();
+  @Output() goalChange = new EventEmitter<Flight>();
```
cavis, line 37: Succinct - i like it!  Will also scale if we add other "goal-ish" fields over there in the future.

cavis (approved):
:+10: the two main changes (daily_min onto flight, and id into flight-form) really clean up the codebase.  And bonus points: bugs fixed!

Did a bit of QA locally - confirmed daily_min is being previewed and persisted (though not a particularly useful setting yet).  Also, couldn't make the flight-tab-switching bug happen, so the id in form appears to have fixed all that.  The advertiser thing still happens for me though ... I'll document that in another more-specific bug ticket!

PR #182, opened 2020-04-24
Fix flight form emit

sandikbarr:
* Bandaid on the Flight Form
  * The boolean guard in wrapping updateFlightForm was no longer doing it. On form valueChanges, I added a check on the fields to see if they changed before emit. 
  * This would _almost do it, but I still saw some updates where the form wasn't dirty that cleared the flight changed status. So there's another fix in the flight.reducer to have changed: false until true on form updates. Changed will be cleared/false on Save.
* Uses the Material Datepicker and uses the MomentDateAdapter with UTC support.
  * Turns out the MomentDateAdapter must be configured in the app.module or it won't take effect
  * This changes the Flight startAt/endAt fields to be Moment instead of native Date. I've rethought this and wondered if I should have just kept Moment in the form, but oh well.
* Cleaned up the Add and Duplicate Flight actions to generate flightId/timestamp in the action creator. The effects don't have a subsequent action and just handle nagivation.
* Spotted a bug where temporary/unsaved flights were not removed on Save if they were softDeleted
* Fixed another bug where duplicated flights were getting an error from availability because it was sending the temp flightId. It's expected that unsaved flights should not have createdAt, fixed.

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -9,6 +9,14 @@ import { ErrorService } from './error/error.service';
 import { AppRoutingModule, routingComponents } from './app.routing.module';
 import { AppComponent } from './app.component';
 
+import { DateAdapter, MAT_DATE_LOCALE, MAT_DATE_FORMATS } from '@angular/material';
+import {
+  MatMomentDateModule,
+  MomentDateAdapter,
+  MAT_MOMENT_DATE_ADAPTER_OPTIONS,
+  MAT_MOMENT_DATE_FORMATS
+} from '@angular/material-moment-adapter';
+
 import { CoreModule } from './core/core.module';
 import { SharedModule } from './shared/shared.module';
 import { CampaignModule } from './campaign/campaign.module';
@@ -34,13 +42,19 @@ import { environment } from '../environments/environment';
     CampaignModule,
     BrowserAnimationsModule,
     DashboardModule,
+    MatMomentDateModule,
     StoreModule.forRoot(reducers, { metaReducers }),
     StoreModule.forRoot({ router: routerReducer }),
     StoreRouterConnectingModule.forRoot({ serializer: CustomRouterSerializer }),
     EffectsModule.forRoot([]),
     !environment.production ? StoreDevtoolsModule.instrument() : []
   ],
-  providers: [{ provide: ErrorHandler, useClass: ErrorService }],
+  providers: [
+    { provide: ErrorHandler, useClass: ErrorService },
+    { provide: DateAdapter, useClass: MomentDateAdapter, deps: [MAT_DATE_LOCALE, MAT_MOMENT_DATE_ADAPTER_OPTIONS] },
+    { provide: MAT_MOMENT_DATE_ADAPTER_OPTIONS, useValue: { useUtc: true } },
+    { provide: MAT_DATE_FORMATS, useValue: MAT_MOMENT_DATE_FORMATS }
+  ],
   bootstrap: [AppComponent]
 })
 export class AppModule {}
--- a/src/app/campaign/flight/flight.component.ts
+++ b/src/app/campaign/flight/flight.component.ts
@@ -1,7 +1,7 @@
 import { Component, OnInit, Input, Output, EventEmitter, ChangeDetectionStrategy } from '@angular/core';
-import { Inventory, InventoryZone } from '../../core';
-import { Flight } from '../store/models';
 import { FormBuilder, Validators, FormArray, AbstractControl } from '@angular/forms';
+import { Inventory, InventoryZone } from '../../core';
+import { Flight, isNameChanged, isInventoryChanged, isStartAtChanged, isEndAtChanged, isZonesChanged } from '../store/models';
 
 @Component({
   selector: 'grove-flight',
@@ -48,7 +48,8 @@ export class FlightComponent implements OnInit {
     startAt: ['', Validators.required],
     endAt: ['', Validators.required],
     zones: this.fb.array([]),
-    set_inventory_uri: ['', Validators.required]
+    set_inventory_uri: ['', Validators.required],
+    totalGoal: [0, Validators.required]
   });
 
   get name() {
@@ -67,7 +68,7 @@ export class FlightComponent implements OnInit {
 
   ngOnInit() {
     this.flightForm.valueChanges.subscribe(flightFormModel => {
-      this.formStatusChanged(flightFormModel, this.flightForm.dirty);
+      this.formStatusChanged(flightFormModel);
     });
   }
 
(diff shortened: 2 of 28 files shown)


src/app/app.module.ts
```diff
@@ -34,13 +42,19 @@ import { environment } from '../environments/environment';
...
     DashboardModule,
+    MatMomentDateModule,
     StoreModule.forRoot(reducers, { metaReducers }),
     StoreModule.forRoot({ router: routerReducer }),
     StoreRouterConnectingModule.forRoot({ serializer: CustomRouterSerializer }),
     EffectsModule.forRoot([]),
     !environment.production ? StoreDevtoolsModule.instrument() : []
   ],
-  providers: [{ provide: ErrorHandler, useClass: ErrorService }],
+  providers: [
```
sandikbarr, line 52: I guess DataAdapters don't work unless applied in the app.module. I had this in the campaign.module when I had tried this before, and that's why it wasn't working. =/

src/app/campaign/flight/flight.component.ts
```diff
@@ -87,32 +88,25 @@ export class FlightComponent implements OnInit {
...
-    // these dates pickers are not actually connected to the reactive form
-    // so in order for the flightForm to be valid on emit, these values must be set on the form
-    if (startAt) {
-      this.flightForm.get('startAt').setValue(startAt, { emitEvent: false });
-    }
-    if (endAt) {
-      this.flightForm.get('endAt').setValue(endAt, { emitEvent: false });
+  // emits updates when reactive form fields change
+  formStatusChanged(flight: Flight) {
+    if (this.isFlightFormChanged(flight)) {
```
sandikbarr, line 103: Even the boolean guard on updateFlightForm was failing me, had to move to checking that the fields had changed.
I'll be looking to do the refactor on this with a parent form and embedded form components next.

src/app/campaign/nav/campaign-nav.component.ts
```diff
@@ -44,10 +44,15 @@ export class CampaignNavComponent {
...
 
-  get dupCampaignState(): { campaign: Campaign; flights: Flight[] } {
+  get dupCampaignState(): { campaign; flights } {
     return {
       campaign: this.campaign,
-      flights: this.flights.filter(flight => !flight.softDeleted).map(flight => flight.localFlight)
+      flights: this.flights
+        .filter(flight => !flight.softDeleted)
+        .map(flight => {
+          // NOTE: router does not want to serialize Moment, so passing thru valueOf, startAt/endAt are removed in reducer
```
sandikbarr, line 53: annoying, but we are dropping these date on dup campaigns anyway

src/app/campaign/store/actions/campaign-action.creator.ts
```diff
@@ -85,25 +88,39 @@ export class CampaignSaveFailure implements Action {
 export class CampaignAddFlight implements Action {
   readonly type = ActionTypes.CAMPAIGN_ADD_FLIGHT;
 
-  constructor(public payload: { campaignId: number | string }) {}
-}
-
```
sandikbarr, line 90: Removed the additional action dispatch from the effects for add and dup flight. The flightId/timestamps are generated in the action creators. Now the effect just handles router navigation without an extra action dispatch, and the reducer responds to the original action.

src/app/campaign/store/models/flight.models.ts
```diff
@@ -55,3 +56,24 @@ export const duplicateFlight = (flight: Flight, tempId: number): Flight => {
 export const getFlightId = (state: FlightState) => {
   return state.localFlight && state.localFlight.id;
 };
+
```
sandikbarr, line 59: util functions to destructure the model and check properties for equality

src/app/campaign/store/reducers/flight.reducer.ts
```diff
@@ -98,7 +98,10 @@ export function reducer(state = initialState, action: CampaignActions): State {
           id: flight.id,
           changes: {
             localFlight,
-            changed,
```
sandikbarr, line 101: Some of the flight form update events had the form.dirty: false, so I decided that with form updates, changed for each flight should be false until true and then remain true until "Save".

src/app/campaign/store/reducers/flight.reducer.ts
```diff
@@ -115,6 +118,12 @@ export function reducer(state = initialState, action: CampaignActions): State {
         state
       );
     }
+    case ActionTypes.CAMPAIGN_SAVE: {
```
sandikbarr, line 121: This was just a bug I noticed where Save wasn't handling temporary flights that we set to softDeleted.

src/app/campaign/flight/flight.component.ts
cavis: +1 makes sense.

src/app/campaign/flight/flight.component.ts
```diff
@@ -87,32 +88,25 @@ export class FlightComponent implements OnInit {
...
-  // emits updates when reactive form fields change
-  formStatusChanged(flight: Flight, changed?: boolean) {
-    this.flightUpdate.emit({
-      flight: { ...flight, totalGoal: this.flight.totalGoal, dailyMinimum: this.flight.dailyMinimum },
-      changed,
-      valid: this.flightForm.valid
-    });
+  isFlightFormChanged(flight: Flight) {
+    return (
+      isNameChanged(flight, this.flight) ||
```
cavis, line 93: Ah, yeah, this is basically what i was thinking with checking json local/remote flights for changes.  Explicit checks work!  (Though would be great if this wasn't necessary).

src/app/campaign/store/actions/campaign-action.creator.ts
```diff
@@ -85,25 +88,39 @@ export class CampaignSaveFailure implements Action {
...
+  public payload: { campaignId: number | string; flightId: number; startAt: Moment; endAt: Moment };
+
+  constructor(payload: { campaignId: number | string; flightId?: number; startAt?: Moment; endAt?: Moment }) {
+    const date = utc();
+    this.payload = {
+      campaignId: payload.campaignId,
+      flightId: payload.flightId || date.valueOf(),
+      startAt:
+        payload.startAt ||
+        utc(date.valueOf())
```
cavis, line 100: Ah, the elusive explicit UTC dates in javascript.  +2.

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -58,30 +58,24 @@ export class CampaignActionService implements OnDestroy {
 
   haveAvailabilityParamsChanged(a: Flight, b: Flight) {
     return (
-      this.hasAvailabilityParams(a) &&
-      (!this.hasAvailabilityParams(b) ||
-        a.startAt.getTime() !== b.startAt.getTime() ||
-        a.endAt.getTime() !== b.endAt.getTime() ||
-        a.set_inventory_uri !== b.set_inventory_uri ||
-        this.haveFlightZonesChanged(a, b))
+      this.hasAvailabilityParams(a) && (isStartAtChanged(a, b) || isEndAtChanged(a, b) || isInventoryChanged(a, b) || isZonesChanged(a, b))
```
cavis, line 61: Seems like this may all go away with the new flight.days API change.

src/app/campaign/store/reducers/flight.reducer.ts
cavis: +1 makes sense.  Since we don't allow "discard" anyways ... may as well keep it marked changed until save.

cavis (approved):
👍 Looks good to me.  momentjs ++.  Nice to have a real way to deal with UTC javascript dates.

cavis:
@sandikbarr i'll leave it to you to merge

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: I thinking this is where it POSTs to the preview endpoint instead

Issue #137, opened 2020-02-28
Implement Flight Inventory Table changes


brandonhundt:
This is a fix for the flights in progress issue. If this feels too large for one ticket, feel free to break it up.

Sketch file: https://sketch.cloud/s/WlAj8/a/L81O73

![Screen Shot 2020-02-27 at 1.15.18 PM.png](https://images.zenhubusercontent.com/5bd1dc3a4cadef67d2695e38/17d536a5-e266-4b57-ad6d-1f4e8ca7e4bb)

- [x] Roll up all zones into a single table (TODO in #170)
- [ ] Status "chip". Text reflects the flight progress (Hold, In progress, Cancelled, etc) Background color represents how well the flight is performing. 

Availability:
- [x] Zero out when the date has passed. Totals include only the remaining values.

Allocated:
- [x] Rename Goal to Allocated
- [x] Dates in the past are bolded and light gray
- [x] If showing preview, add `*`

Actuals:
- [x] Dates in the past are bolded and light gray

Color Notes:
- [ ] Orange dashed border line represents a week with an episode drop (If expanded this should show update to the day) -- If challenging I can take or leave it (TODO in #169)
- [x] Blue border line represents the current date (If expanded, should follow the current date)
- [x] Use red text if the flight is borked and requires manual intervention (TODO in #168)

Expanded:
[Screen Shot 2020-02-27 at 1.57.27 PM.png](https://images.zenhubusercontent.com/5bd1dc3a4cadef67d2695e38/e9a99267-1d69-454b-bfb1-9ce688de0a66)

sandikbarr:
@brandonhundt These are Campaign Status values:
| Status "chip". Text reflects the flight progress (Hold, In progress, Cancelled, etc)
You want this campaign status shown on each flight's inventory table? Or did you want the flight status ("Ok" or "Error")?

sandikbarr:
To show some sort of indicator when the unsaved preview is shown as opposed to the allocated amount, it uses the color green and an asterisk(*) on the allocations column. How should preview be indicated in the new design?

sandikbarr:
> Use red text if the flight is borked and requires manual intervention

What does this mean more mathematical terms?
Currently, a row would be highlighted red if the allocation preview > allocated + available, or if on allocation preview the Allocated column > Available column.

brandonhundt:
[...]

PR #186, opened 2020-04-28
inventory table changes

sandikbarr:
Some of the changes requested in #137
  - When the inventory day's date has passed, zero out availability and display data as bolded in the color light grey
  - When showing the allocation preview data, the column is marked with an `*` and is no longer colored green
  - A blue border line represents the current date. If a week containing the current date is not expanded, the current date indicator is shown on the weekly row. If the week is expanded, the indicator line is shown on the expanded daily row.

Additionally, I have refactored some of the selector code that creates these rollups based on feedback of the really long function that I moved from the CampaignStateService and plopped down into a selector function during the refactor. No one likes to doing bunch of data manipulation and transformations in JavaScript, but I do hope it is more readable. I also moved summing the available and allocated numbers into the selector and out of the component as well as implemented the zeroing of availability data prior to the current date in the selector.

### DIFF
--- a/src/app/campaign/availability/availability.component.scss
+++ b/src/app/campaign/availability/availability.component.scss
@@ -34,6 +34,16 @@ section {
     border-top: $row-divider-border;
   }
 
+  & + &.current-date,
+  &.current-date {
+    border-top: 1px solid mat-color($prx-primary, lighter);
+  }
+
+  &.past-date * {
+    color: mat-color($prx-neutral, default);
+    font-weight: 700;
+  }
+
   &.head {
     background-color: prx-theme-background(table-header);
 
@@ -139,8 +149,6 @@ section {
 }
 
 .preview {
-  color: $success;
-
   &::after {
     content: '*';
   }
--- a/src/app/campaign/availability/availability.component.spec.ts
+++ b/src/app/campaign/availability/availability.component.spec.ts
@@ -173,10 +173,6 @@ describe('AvailabilityComponent', () => {
     expect(comp.getAllocationValue(week.numbers)).toEqual(week.numbers.allocationPreview);
   });
 
-  it('should return combined value of allocated and availability', () => {
-    expect(comp.getAvailableValue(week.numbers)).toEqual(week.numbers.allocated + week.numbers.availability);
-  });
-
   it('should show a week as the current week when not expanded', () => {
     expect(comp.isZoneWeekExpanded(mockRollup, week)).toBeFalsy();
     expect(comp.showAsCurrentWeek(mockRollup, week)).toBeFalsy();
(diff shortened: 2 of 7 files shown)


src/app/campaign/availability/availability.component.scss
```diff
@@ -34,6 +34,16 @@ section {
     border-top: $row-divider-border;
   }
```
sandikbarr, line 36: Both `&.` and `& + &.` rules because we have an `& + & rule` with a border-top I needed to override, aint specificity great

src/app/campaign/availability/availability.component.spec.ts
```diff
@@ -24,31 +24,31 @@ describe('AvailabilityComponent', () => {
       {
         startDate: new Date('2019-10-01'),
         endDate: new Date('2019-10-05'),
```
sandikbarr, line 26: The component now expects that the availability numbers include the allocated amount. Previously the rollups were just doing weekly and totals and letting the component display availability as allocated + availability

src/app/campaign/store/selectors/availability-allocation-actuals.selectors.ts
```diff
@@ -1,9 +1,141 @@
...
+  if (allocationPreviewEntities && allocationPreviewEntities[`${zone}_${dateString}`]) {
+    return allocationPreviewEntities[`${zone}_${dateString}`].goalCount;
+  }
+};
+
+export const getMidnightUTC = (date: Date): number => {
+  return date.valueOf() - (date.valueOf() % (24 * 60 * 60 * 1000));
+};
+
+export const getEndOfWeek = (date: Date): Date => {
```
sandikbarr, line 29: yeah... so there's a mix of Date and Moment in this because of the timing on the whole thing of adding moment into the project while I was in progress on this ticket. Moment is a bit less heavy lifting with its UTC mode.

cavis (approved):
👍 Changes look good to me.  Did hit a couple weird data issues and bugs while testing this ... but they were all pre-existing.

The bug I saw that's in Grove: the "allocated" column doesn't reload after you save a flight.  It's actually reverting to the previous counts.  So if I change the total goal, the preview numbers show that change.  Then hitting save, it reverts.

I'm guessing this will just be solved by the API change (since all the numbers are coming back with the flight json itself) ... so not worth investing time into it here!

sandikbarr:
Huh, yeah I'm not sure it's ever been actually reloading the new availability on save. In my head I thought it did but I don't see where it was supposed to be doing this in the code. You're correct, the API changes should make this easier to fix.
* edit: you know, if that ever worked after Save I probably broke it in the refactor to NgRx

PR #187, opened 2020-05-07
181: delete campaign

### LINKED ISSUE #181: Ability to delete a campaign
- [ ] Add a button under duplicate campaign to delete it. Styling red dashed outline. Red text. Trash icon.
- [ ] A campaign with flights that have served impressions (actuals) cannot be deleted.
- [ ] After its clicked, ensure there's an alert confirming this is actually what the user wants to do.

rpeterman-gp:
Closes #181 

### Changes
- add menu to campaign status bar
- add delete button to menu
- add duplicate button to menu
- remove duplicate button from nav
- add state management for campaign delete
- refactor campaign duplicate trigger process
- various style tweaks to fix some `::ng-deep styling`

## Things to tests
- [ ] Delete a campaign from campaign status bar menu.
   - Button should NOT be available for campaigns with actuals.
   - Button should be disabled while saving.
   - Should be get confirmation prompt before delete occurs.
- [ ] Duplicate a campaign from campaign status bar menu.
  - Button should enable/disable under the same conditions as the previous button in the nav.
  - Duplicated campaign and flights now have "Copy of " prepended to their names. Should make it clearer than it is a duplicate and encourage user to update the name.


src/app/campaign/store/effects/campaign.effects.ts
```diff
@@ -151,11 +152,47 @@ export class CampaignEffects {
...
+        flights: flights.map(flight => ({
+          ...flight,
+          startAt: undefined,
+          endAt: undefined
+        }))
+      };
+      this.router.navigate(['/campaign', 'new'], { state });
+    })
+  );
+
```
sandikbarr, line 171: alternatively, I feel like the one effect for action CAMPAIGN_DUP_FROM_FORM could have done the router.navigate as well as the toastr.success. It's a weird one since it's using effects but actually synchronous since we don't wait on the navigation or anything and just expect that the app reactively works when these things happen, so I'm really just mentioning this as a thought.

sandikbarr (approved):
Looks great! I just made a couple changes for unused imports, format on save, nothing functional. Glad you got that test, I grabbed it too but hadn't pushed yet, so sorry for the extra noise. Ready to be merged when the tests pass/build finishes.

PR #192, opened 2020-05-11
flight.days and preview api migration

sandikbarr:
Migrates away from allocation preview and inventory availability APIs to the new flight.days and preview APIs
  - flight.days stored on campaign load and campaign save
  - flight preview loaded on flight form (except name field) or goal form changes
  - flight preview and preview error are stored in the same store as flight.days and overwrite the flight.days data when loaded (which might be problematic if we do a form discard changes, but we'll reshape the data at that point as needed)


src/app/campaign/inventory/inventory-table.component.ts
```diff
@@ -0,0 +1,122 @@
+import { Component, ChangeDetectionStrategy, Input } from '@angular/core';
+import { Flight, FlightDay, InventoryRollup, InventoryWeeklyRollup } from '../store/models';
+import { getMidnightUTC, getDateSlice } from '../store/selectors';
+
```
sandikbarr, line 4: Extracted from the previously named availability.component and renamed to inventory-table. I felt like this table part needed to be its own component.

src/app/campaign/store/actions/campaign-action.service.spec.ts
```diff
@@ -45,48 +52,48 @@ describe('CampaignActionService', () => {
...
-        flightId,
-        createdAt,
-        set_inventory_uri,
-        name,
-        startAt: startAt.toDate(),
-        endAt: endAt.toDate(),
-        totalGoal,
-        dailyMinimum,
-        zones
-      })
```
sandikbarr, line 89: jest seems to have a matcher bug where it fails with "values have no visual difference," so I resorted to using JSON.stringify on the spy.mock.calls instead of toHaveBeenCalledWith. annoying.

src/app/core/campaign/flight-preview.service.ts
```diff
@@ -0,0 +1,46 @@
...
+      startAt: string;
+      endAt: string;
+      totalGoal: number;
+      dailyMinimum: number;
+      zones: string[];
+    },
+    flightDoc?: HalDoc,
+    campaignDoc?: HalDoc
+  ): Observable<HalDoc[]> {
+    if (flightDoc) {
```
sandikbarr, line 36: POST is giving a 404 on these first two, but it's working on the last one on the api root

src/app/campaign/inventory/inventory-table.component.ts
cavis: +10.

src/app/campaign/inventory/inventory-table.component.ts
```diff
@@ -0,0 +1,122 @@
...
+        <div>{{ week.numbers.available | largeNumber }}</div>
+        <div [class.preview]="isPreview">
+          {{ week.numbers.allocated | largeNumber }}
+        </div>
+        <div>
+          {{ showActualsValue(week.startDate) ? (week.numbers.actuals | largeNumber) : '&mdash;' }}
+        </div>
+        <div class="edit">
+          <button class="btn-link" aria-label="Edit">
+            <prx-icon name="pencil" size="14px" color="primary"></prx-icon>
```
cavis, line 41: I do wonder: if/when we get around to making these manually editable/lockable, how this component will interact with the form changed/valid state.

A problem for another day!

src/app/campaign/inventory/inventory.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  @Input() isPreview: boolean;
+  @Input() previewError: any;
+  @Output() goalChange = new EventEmitter<Flight>();
+  zoneWeekExpanded = {};
+
+  get errors() {
+    return [
+      // allocation preview error
+      // TODO: Updated with discussed "nice_message" when available.
+      this.previewError && this.previewError.body && this.previewError.body.message,
```
cavis, line 41: Just noting: this will go away per PRX/augury.prx.org#233.  Previews will _always_ return success (unless something _really_ goes wrong).  And you'll have to check the returned flight `status` / `status_message` to see if we actually met your total_goal.

src/app/campaign/store/actions/campaign-action.service.spec.ts
cavis: Vanilla JS wins again!

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -1,106 +1,62 @@
...
-
-  loadAvailabilityAllocationIfChanged(formFlight: Flight, localFlight: Flight) {
-    if (this.isDateRangeValid(formFlight) && this.haveAvailabilityParamsChanged(formFlight, localFlight)) {
-      this.loadAvailability({ ...formFlight, id: localFlight.id, createdAt: localFlight.createdAt });
-      if (formFlight.totalGoal) {
-        this.loadAllocationPreview({ ...formFlight, id: localFlight.id, createdAt: localFlight.createdAt });
-      }
-    }
+  loadFlightPreview(flight: Flight) {
+    const { id: flightId, set_inventory_uri, name, startAt, endAt, zones, totalGoal, dailyMinimum } = flight;
```
cavis: In theory ... couldn't you just pass _all_ of the flight data through to the preview POST?  It should accept all the same params - it's actually doing a create/update (POST/PUT) behind the scenes, and rolling back the db transaction before returning the json.

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -1,106 +1,62 @@
...
+    this.store.pipe(select(selectRoutedCampaignFlightDocs), first()).subscribe(({ campaignDoc, flightDoc }) => {
+      const params = {
+        name,
+        set_inventory_uri,
+        startAt: startAt.toDate(),
+        endAt: endAt.toDate(),
+        totalGoal: totalGoal || 0,
+        dailyMinimum: dailyMinimum || 0,
+        zones: zones.map(zone => zone.id)
```
cavis: Yeah, zones can be saved as-is now.  May be enough to just pass through the Flight model.

src/app/campaign/store/effects/campaign.effects.ts
```diff
@@ -83,6 +89,70 @@ export class CampaignEffects {
...
+                campaignDoc,
+                deletedFlightDocs,
+                updatedFlightDocs,
+                createdFlightDocs,
+                createdFlightDaysDocs
+              }))
+            );
+          }
+        ),
+        // retrieve the flights' days from each updated flight response
```
cavis, line 112: Comments++!  (And some good observable-fu loading all the campaign -> flights -> flight-days in here!)

src/app/campaign/store/models/flight-days.models.ts
```diff
@@ -0,0 +1,57 @@
+import { utc } from 'moment';
+import { HalDoc } from 'ngx-prx-styleguide';
+
+export interface FlightPreviewParams {
```
cavis: (It should be enough to just use the `Flight` for this - one less thing to keep in-sync).

src/app/core/campaign/campaign.service.ts
```diff
@@ -33,6 +33,31 @@ export class CampaignService {
     );
   }
 
+  loadCampaignZoomFlightsAndFlightDays(
+    id: number
+  ): Observable<{ campaignDoc: HalDoc; flightDocs: HalDoc[]; flightDaysDocs: { [id: number]: HalDoc[] } }> {
+    return this.augury.follow('prx:campaign', { id, zoom: 'prx:flights,prx:flight-days' }).pipe(
+      switchMap(campaignDoc => campaignDoc.follow('prx:flights').pipe(map(flightDocs => ({ campaignDoc, flightDocs })))),
+      switchMap(({ campaignDoc, flightDocs }: { campaignDoc: HalDoc; flightDocs: HalDoc }) => {
+        // if total is greater than count, request all flights
```
cavis, line 42: Ooh, nice edge case catching.

cavis (approved):
Woooo!  I follow some of these things!  Very helpful comments on the longer pipes.  And seems like some good test improvements.

Left a few comments on how future Augury tickets may impact bits of this code.

And one suggestion on getting rid of the flight-preview-params model.  (It _is_ the `Flight` form model).  Just so we're not maintaining 2 duplicate things.

But totally a :+1.4:

src/app/campaign/inventory/inventory.component.ts
sandikbarr: ok cool, it also displays any status message in the inventory.component. maybe we'll never have to do that error formatting/handling after all. agile!

src/app/campaign/inventory/inventory-table.component.ts
sandikbarr: I feel like we can just send that, back up and down the state. Hopefully the next reactive forms refactor will make that form/model interaction more better.

src/app/campaign/inventory/inventory.component.ts
sandikbarr: Even though it's going away, I found a bug in this preview error as well where I don't clear itself within the campaign because the state doesn't change on flight route change anymore, small fix incoming to track those errors in the store by flightId.

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: you're correct on that. I had carried the params forward unnecessarily. Refactored! I did choose to keep the Flight model itself out of the flight-preview.service because it's in core.module not in the campaign.module (even though I have not been entirely consistent about that, it's aspirational ok)

src/app/campaign/store/effects/campaign.effects.ts
sandikbarr: positive reinforcement++ it works!

sandikbarr:
Refactored and removed the the FlightPreviewParams and adjusted how the flight previewError was kept in state, think it's in good shape now.

cavis (approved):
👍 yeah!

Issue #92, opened 2019-12-10
Implement Uncapped Flights


brandonhundt:
Needs PRX/augury.prx.org#173.

* Lower priority than capped flights (uncapped flights = you get what you get)
* Warn users that there’s an uncapped flight also targeted, and that capped flight will eat into that inventory. (status not warning). Can be displayed under the table.
* Use Material's slide toggle: https://material.angular.io/components/slide-toggle/overview
* BONUS: Clean up spacing around goal fields.
* If toggled, the goal/minimum fields will not be present.
* If toggled, the “Allocated” column in the table changed to “Expected”, and it would display the day over day counts of “available_count - SUM(other_allocations)”

![cap-toggle](https://user-images.githubusercontent.com/369739/70573716-27699100-1b68-11ea-9bed-6ccb600d76e2.jpg)

sandikbarr:
Not sure the API lets you save a flight without a goal at this point, maybe if uncapped, will have to check.

cavis:
Checking the Augury code... looks like there's a ["uncapped"](https://github.com/PRX/augury.prx.org/blob/master/app/representers/api/flight_representer.rb#L15) param on the flight now.  And if you set that to true, it validates that the [goal must be 0](https://github.com/PRX/augury.prx.org/blob/master/app/models/flight.rb#L137).

sandikbarr:
Is it cool if I pull the other flight allocations from the `/inventory/availability` API? That's where allocations come from until the API refactoring is done and we can move all the inventory numbers to be off the flights.

cavis:
I think so.  We still want to display availability and actuals even if your daily allocations are 0.

I'd assume with that API refactoring (PRX/augury.prx.org#205), you'll basically get the uncapped flight as:

```
{
  name: "My Flight Name",
  totalGoal: 0,
  uncapped: true,
  days: [
    {date: '2020-02-22', actuals: 123, allocated: 0, availability: 789}
    ...
   ]
}
```
```

sandikbarr:
nice. Since those are embedded in flights we probably shouldn't even need to zoom all the flights and the numbers to calculate the uncapped flight should just be available from the original campaign + flights request. thank you!

cavis:
I may try to pull this in this sprint, if i can.

PR #194, opened 2020-05-13
Uncapped flights

cavis:
For #92.

Simplistic take on "uncapped flights" - will likely change from a boolean to an enum field in PRX/augury.prx.org#226.  (We'll need a frontend ticket out of that).

But for now, just set the `uncapped = true` field, which reduces allocations to 0.  And display a slightly different table:

image


src/app/campaign/campaign.module.ts
```diff
@@ -5,6 +5,7 @@ import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatDatepickerModule } from '@angular/material/datepicker';
 import { MatInputModule } from '@angular/material/input';
 import { MatSelectModule } from '@angular/material/select';
+import { MatSlideToggleModule } from '@angular/material/slide-toggle';
```
cavis, line 8: Using this introduced a bunch of warnings into the tests.

```
console.warn node_modules/@angular/core/bundles/core.umd.js:16879
  The "slidestart" event cannot be bound because Hammer.JS is not loaded and no custom loader has been specified.
console.warn node_modules/@angular/core/bundles/core.umd.js:16879
  The "slide" event cannot be bound because Hammer.JS is not loaded and no custom loader has been specified.
```

Had to include hammerjs in jest to get rid of them.

src/app/campaign/flight/flight.component.ts
```diff
@@ -48,8 +48,7 @@ export class FlightComponent implements OnInit {
     startAt: ['', Validators.required],
     endAt: ['', Validators.required],
     zones: this.fb.array([]),
-    set_inventory_uri: ['', Validators.required],
-    totalGoal: [0, Validators.required]
```
cavis, line 52: I removed this because the `flight.component.html` didn't have the field.  But apparently that's not necessary - and the spec was testing this.

So... (A) is this still needed, or leftover - and (B) if it is needed, shouldn't dailyMin be in here too?

src/app/campaign/inventory/goal-form.component.scss
```diff
@@ -12,4 +12,20 @@ $gap-size: 0.75rem;
...
+  line-height: 2rem;
+}
+
+.warn {
+  color: $warn;
+  font-weight: 700;
+}
+
+mat-slide-toggle {
+  justify-self: right;
```
cavis, line 29: I wish i knew how to switch this back to justify-left when the grid wraps to 1-column-wide.

src/app/campaign/inventory/inventory-table.component.ts
```diff
@@ -8,8 +8,17 @@ import { getMidnightUTC, getDateSlice } from '../store/selectors';
     <div class="row head">
       <div class="expand"></div>
       <div class="date">Week</div>
-      <div class="avail">Available</div>
-      <div class="goal">Allocated</div>
+
+      <ng-template [ngIf]="uncapped" [ngIfElse]="capped">
```
cavis: Had a ridiculously hard time testing that changing the `comp.flight.uncapped` value changed these headers.  In fact, when I looked at the `de.nativeElement.innerHTML` content, _neither_ "Expected" nor "Available" was in there.  Just a bunch of native angular bindings - but the text in these ng-templates wasn't actually rendered?  Arggggg.

So I gave up.

src/app/campaign/inventory/inventory.component.ts
```diff
@@ -64,7 +64,7 @@ export class InventoryComponent {
       !this.flight.startAt ||
       !this.flight.endAt ||
       !this.flight.set_inventory_uri ||
-      !(this.flight.zones && this.flight.zones.length)
+      !(this.flight.zones && this.flight.zones.filter(z => z.id).length)
```
cavis, line 67: Caught a bug here ... for a new flight, zones is actually `[{id: null, label: null, url: null}]`

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -95,8 +95,10 @@ export class CampaignActionService {
   }
 
   setFlightGoal(flight: Flight) {
-    const { id, totalGoal, dailyMinimum } = flight;
-    this.store.dispatch(new campaignActions.CampaignFlightSetGoal({ flightId: id, totalGoal, dailyMinimum, valid: !!totalGoal }));
+    const { id, totalGoal, dailyMinimum, uncapped } = flight;
+    // TODO: goals of 0 seem to be fine, even for capped flights
+    const valid = totalGoal >= 0 && dailyMinimum >= 0;
```
cavis, line 99: Was thinking this through in the context of PRX/augury.prx.org#233 ... we don't actually need to prevent them from saving a goal of 0 (as a placeholder flight or something).  Augury already allows a goal of 0.

(And uncapped flights always have goal/daily-min of 0).

src/app/campaign/inventory/goal-form.component.ts
```diff
@@ -37,23 +50,32 @@ export class GoalFormComponent implements OnInit, OnDestroy {
...
         debounceTime(200),
         distinctUntilChanged((a, b) => {
-          return a.totalGoal === b.totalGoal && a.dailyMinimum === b.dailyMinimum;
+          return a.totalGoal === b.totalGoal && a.dailyMinimum === b.dailyMinimum && a.uncapped === b.uncapped;
         })
       )
       .subscribe(cmp => {
-        this.formStatusChanged(cmp);
+        if (cmp.uncapped) {
+          this.formStatusChanged({ ...cmp, totalGoal: 0, dailyMinimum: 0 });
```
cavis, line 75: This is where I "reset" the goals when the user flips to "uncapped".  Let me know if it's an ngrx party foul, or if there's a better way to accomplish this.

src/app/campaign/flight/flight.component.ts
sandikbarr: I think I was trying to make sure it wasn't dropping fields, but let's assume it's not needed. The new forms _should_ fix all of this.

src/app/campaign/inventory/goal-form.component.ts
sandikbarr: I think it's probably fine because it's not _supposed_ to emitEvent and we've got that extra guard on form changes...
Might suggest another way of doing it with reactive forms and rxjs by separating out the observables to each of the form component's changes. I can experiment with that as I get these forms broken up.

src/app/campaign/inventory/inventory-table.component.ts
sandikbarr: Ugh, looked into this, and there is definitely a problem with template re-rendering here, but it doesn't seem _entirely_ like our bug. It didn't seem like component would re-render just by setting `comp.flight.uncapped = true` because the Object reference does not change by setting a property, and change detection does not run on deep object property changes. So I changed that to `comp.flight = {..comp.flight, uncapped: true}`, nope still seeing the Allocated heading.  Then I found this https://github.com/angular/angular/issues/12313 and tried adding a ParentComponent wrapper in the test, no luck. I also give up on this test.

You do have one problem though, you're re-using the #capped template ref, so in my test I was actually see the numeric values in the second #capped template ref. Those need to be unique.

sandikbarr (changes requested):
Rename one of the #uncapped template refs in inventory-table. Those need to be unique. Otherwise, good to go.

src/app/campaign/inventory/inventory-table.component.ts
cavis: Good to know I'm not crazy.  Or that we're both equally crazy.

Will fix those refs!

src/app/campaign/inventory/goal-form.component.ts
cavis: +1 the more examples of correctly using reactive-forms + rxjs, the better for us all!

sandikbarr (approved):
👍

PR #196, opened 2020-05-18
fixes loading campaign with no flights

sandikbarr:
An empty forkJoin completes immediately and does not emit. So if the flights array is empty, return an Observable of an empty Object for flight.days instead.


cavis (approved):
👍

PR #200, opened 2020-05-21
Flight Form ControlContainer

### LINKED ISSUE #184: Flight Form: Parent Form with Nested Form Components
Refactor the flight and availability forms as nested form components under a single parent form container using Angular's ControlValueAccessor to embed the nested form components.

sandikbarr:
closes #184


src/app/campaign/flight/flight-form-control-container.component.ts
```diff
@@ -0,0 +1,168 @@
+import { Component, OnInit, OnDestroy, Input, Output, EventEmitter, ChangeDetectionStrategy } from '@angular/core';
```
sandikbarr, line 1: This ControlContainer component adds another layer to the flight form components. We have the flight.container, which subscribes to the ngrx selectors with async pipes so we don't have to track those subscriptions or store that state data anywhere extra. Following the flight.container is this ControlContainer component that has the parent form in its template and contains components that have form fields. The inner form field components inject ControlContainer so that their form events bubble up to the parent form.

Here's a post on using the ControlContainer with nested forms: https://jenniferwadella.com/blog/nested-angular-reactive-forms

sandikbarr: The validations all have to be set up with the parent form unless child form field components implement the ControlValueAccessor and Validator interfaces to encapsulate those validations. That step is something I would recommend for the more complex fields like zones so we don't have to stick validations up here so far away from the fields they pertain to.
Generally, think of ControlValueAccessor for custom form components. Angular provides values accessors for the standard form components.
This post is about nesting forms with ControlValueAccessor: https://indepth.dev/angular-nested-reactive-forms-using-controlvalueaccessors-cvas/#reason--8/

src/app/campaign/flight/flight-form.component.ts
```diff
@@ -0,0 +1,48 @@
...
+  @Output() flightDeleteToggle = new EventEmitter(true);
+  @Output() addZone = new EventEmitter();
+  @Output() removeZone = new EventEmitter<number>();
+  flightForm: FormGroup;
+
+  ngOnInit() {
+    this.flightForm = this.formContainer.control as FormGroup;
+  }
+
+  constructor(private formContainer: ControlContainer) {}
```
cavis, line 27: The magic!

And I see that the angular docs on `ControlContainer` are ... lacking.

src/app/campaign/flight/flight-form.component.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  inventory = inventoryFixture;
+  zoneOptions = inventoryFixture[0].zones;
+  flight = flightFixture;
+  softDeleted = false;
+  flightDeleteToggle = jest.fn(() => {});
+  flightDuplicate = jest.fn(() => {});
+  addZone = jest.fn(() => {});
+  removeZone = jest.fn(() => {});
+
+  flightForm = this.fb.group({
```
cavis, line 77: So what happens if you don't setup (or leave a field out of) the parent form element?  Does the `ControlContainer` just fail?

src/app/campaign/flight/flight.container.ts
```diff
@@ -85,10 +75,6 @@ export class FlightContainerComponent implements OnInit, OnDestroy {
     this.campaignAction.updateFlightForm(formFlight, changed, valid);
   }
 
-  onGoalChange(flight: Flight) {
-    this.campaignAction.setFlightGoal(flight);
```
cavis, line 89: I wonder what triggers the preview now ... /me reads onwards!

src/app/campaign/inventory/goal-form.component.ts
```diff
@@ -34,71 +31,16 @@ import { Flight } from '../store/models';
...
     return !this.goalForm.get('uncapped').value;
   }
 
-  constructor(private fb: FormBuilder) {}
+  constructor(private formController: ControlContainer) {}
 
   ngOnInit() {
-    this.goalFormSub = this.goalForm.valueChanges
-      .pipe(
-        debounceTime(200),
```
cavis, line 68: Do totalGoal / dailyMinimum changes still get debounced?

src/app/campaign/inventory/goal-form.component.spec.ts
```diff
@@ -1,61 +1,57 @@
...
-
-  it('sets the goals to 0 for uncapped flights', done => {
-    comp.flight = flight;
-    comp.goalChange.subscribe(change => {
-      expect(change).toMatchObject({ ...flight, dailyMinimum: 0, totalGoal: 0, uncapped: true });
-      done();
-    });
+  it('show form controls based on capped/uncapped form field value', () => {
+    comp.goalForm.get('uncapped').setValue(false);
+    expect(de.query(By.css('input[type="number"]')).nativeElement).toBeDefined();
```
cavis, line 51: Ooh, so DOM changes are working in tests now?  What fixed it?

src/app/campaign/inventory/inventory.component.ts
```diff
@@ -10,7 +10,7 @@ import { Flight, InventoryRollup } from '../store/models';
       <p>Please select Start and End Dates, Series, and Zones to view inventory.</p>
     </div>
     <ng-template #inventory>
-      <grove-goal-form [flight]="flight" (goalChange)="goalChange.emit($event)"></grove-goal-form>
+      <grove-goal-form></grove-goal-form>
```
cavis, line 13: Well that's cleaner!

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -15,12 +16,54 @@ import {
...
+  flightFormValueChanges = new Subject<{ flight: Flight; changed: boolean; valid: boolean }>();
+  flightFormValueChangesSubscription: Subscription;
+  flightFormValueChangesStream = this.flightFormValueChanges.pipe(
+    // map form field values to a proper Flight
+    map(({ flight, changed, valid }) => {
+      return {
+        flight: {
+          ...flight,
+          uncapped: flight.uncapped || false,
+          totalGoal: flight.uncapped ? 0 : +flight.totalGoal,
```
cavis, line 30: Are these just a way to setup some default values?

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -15,12 +16,54 @@ import {
...
+        this.loadFlightPreview(formState.flight);
+      }
+    }),
+    // guard flight form updates
+    filter(([formState, flightState]) => {
+      // are we still dealing with these form updates that aren't actual updates?? maybe just on route change at this point
+      const nameChanged = flightState.localFlight.name !== formState.flight.name;
+      const previewParamsChanged = this.havePreviewParamsChanged(flightState.localFlight, formState.flight);
+      if (!nameChanged && !previewParamsChanged) {
+        console.warn('unchanged?', { formState }, { flightState });
```
cavis: I do get this warning everytime I change the TotalGoal or DailyMin fields using the up/down arrows.  Or hitting backspace a bunch of times.

src/app/campaign/store/models/flight-days.models.ts
```diff
@@ -17,6 +17,7 @@ export interface FlightDays {
   flightDoc?: HalDoc;
   flightId: number;
   days: FlightDay[];
+  preview?: boolean;
```
cavis, line 20: Ah!  I was wondering about this - it was at the wrong level of the state before.

src/app/campaign/store/reducers/flight.reducer.ts
```diff
@@ -92,7 +90,12 @@ export function reducer(state = initialState, action: CampaignActions): State {
     }
     case ActionTypes.CAMPAIGN_FLIGHT_FORM_UPDATE: {
       const { flight, changed, valid } = action.payload;
-      const localFlight = { ...(state.entities[flight.id] && state.entities[flight.id].localFlight), ...flight };
+      const localFlight = {
+        ...(state.entities[flight.id] && state.entities[flight.id].localFlight),
+        ...flight,
+        dailyMinimum: flight.dailyMinimum || 0,
```
cavis, line 96: I'm still a bit confused where default values are set - some overlap with the form setting defaults, the reducer, and the actions.  Is there any standard place to do that in reactive forms, or more of an "everybody set their own defaults" belt-and-suspenders thing?

src/app/core/inventory/inventory.service.ts
```diff
@@ -16,6 +16,23 @@ export interface InventoryZone {
   label: string;
 }
 
+// filter zone options to the control at an index
+export const filterZones = (allZones: InventoryZone[], selectedFlightZones: InventoryZone[], zoneIndex?: number) => {
```
cavis, line 20: Wait - didn't I already see this method?

cavis: Yeah, up in `flight-form-control-container.component.ts`.  Should this be in both places, or is one leftover from the refactor?

cavis (changes requested):
:+ten: Nicely done - really helps cleanup those complex forms.  (Which will only get more complex, with pingbacks and flight-targeting).

Couple questions, maybe one suggestion, and one maybe redundant function.

src/app/campaign/inventory/goal-form.component.spec.ts
sandikbarr: I don't know for sure as this should have always worked, but I saw it too when you were having that issue.

src/app/campaign/inventory/goal-form.component.ts
sandikbarr: No. :( We need to implement a custom form component with the ControlValueAccessor in order to control when these fields emit changes. I figured that was a next step. I liked that they debounced, but I'm not sure anyone noticed or will miss it that much? I tried debouncing in that stream, but tbh that made the tests difficult.

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: No, actually this is where it sets totalGoal and dailyMinimum to zero if the flight is uncapped. Also the incoming fields here are from the form off the valueChanges, so they are coerced to number with the `+` prefix.

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: k, seems proof we still need the guard for whatever reason. That stinks because every new form field we add will have to be added to that guard, and they are just getting more complex from here out. 

I think I should remove or comment out this warning.

It does seem like `previewParamsChanged` would be `true` when you use the up/down arrows or backspace because it checks all the form params for changes that aren't name.

Frustrating.

src/app/campaign/store/models/flight-days.models.ts
sandikbarr: fixed. forever. was maybe not my best work, ok

src/app/campaign/inventory/goal-form.component.ts
cavis: +1 that's fine.  Let's get this out "in the wild" ... then get more feedback on form responsiveness.

src/app/campaign/flight/flight.container.ts
cavis: (I found it!)

src/app/campaign/store/actions/campaign-action.service.ts
cavis: Seems like key events emit twice, for some reason?  So the 2nd time, preview params _haven't_ changed (and we're in-progress loading that preview).

src/app/campaign/store/actions/campaign-action.service.ts
cavis: Ohhh, okay. That makes more sense.  (Also resolves some of my "where to set defaults" question).

src/app/campaign/store/reducers/flight.reducer.ts
sandikbarr: There is a standard place for the initial values in reactive forms when you set them up with the form builder, but not defaults as in if a field is not set in a form update, default to this value.

We do have these fields initialized https://github.com/PRX/grove.prx.org/pull/200/files#diff-51fa2200ec802e5c3d1a18c7dc52a7a1R83, but undefined/null/empty values were getting thru.

src/app/core/inventory/inventory.service.ts
sandikbarr: You're not wrong. It's not just because you originally wrote that code either. I meant to move that into one place here and failed to finish that refactoring. Will fix.

src/app/campaign/flight/flight.container.ts
sandikbarr: mystery solved! yeah, it all goes through the single form stream that's piped through the subject to filter those updates to be dispatched out

src/app/campaign/flight/flight-form.component.spec.ts
sandikbarr: If there's a field on the form you didn't set up in the form group, it will fail with error `Cannot find control with name: 'xyz'`

src/app/campaign/flight/flight.container.ts
cavis: yeah, and this seems quite a bit cleaner than the form telling you the goal changed directly.  +5.

src/app/campaign/inventory/inventory.component.ts
sandikbarr: Yes, it's just kind of a way to organize the child templates/form controls at this point. Much better. We might want to use ControlValueAccessors with the more complex fields to keep that code together. But for simple form elements, these are just basically child templates now and just need to handle their own display.

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: That makes me uncomfortable, and I don't like it. It's better and things are going in the right direction, but I suspected this might not solve all of our form problems.

src/app/core/inventory/inventory.service.ts
sandikbarr: looked like I was using this function in the component, just hadnt removed it from the class. fixed.

sandikbarr:
Think I got all the comments addressed. Thanks so much for looking this over and also doing some testing!

src/app/campaign/store/actions/campaign-action.service.ts
cavis: +1.  mo forms, mo problems.

cavis (approved):
👍 looks good to me!

PR #202, opened 2020-05-26
Flight preview status

### LINKED ISSUE #197: Handle allocation status message from preview
Before https://github.com/PRX/augury.prx.org/pull/238

- [ ] Pull error status message from the preview'd flight model.  Currently this is pulled from the saved flight.
- [ ] Clear/redraw error messages when preview succeeds or errors change.  When the user trips an error, then fixes it, the displayed error goes away.

cavis:
closes #197.  Should deploy along with PRX/augury.prx.org#238.  Exceeding availability no longer prevents you from saving the flight - but it does return a `statusMessage`.

This patches through any 201 preview's `status` and `statusMessage` back onto the localFlight so you can see how your pending changes will impact the flight.

image

Also, I kept around the old previewError as that catches/displays some server validation errors nicely:

image

### DIFF
--- a/src/app/campaign/store/reducers/flight.reducer.ts
+++ b/src/app/campaign/store/reducers/flight.reducer.ts
@@ -1,6 +1,7 @@
 import { EntityState, EntityAdapter, createEntityAdapter } from '@ngrx/entity';
 import { ActionTypes } from '../actions/action.types';
 import { CampaignActions } from '../actions/campaign-action.creator';
+import { FlightPreviewActions } from '../actions/flight-preview-action.creator';
 import { docToFlight, duplicateFlight, Flight, FlightState, getFlightId } from '../models';
 
 export interface State extends EntityState<FlightState> {
@@ -12,7 +13,7 @@ export const adapter: EntityAdapter<FlightState> = createEntityAdapter<FlightSta
 
 export const initialState: State = adapter.getInitialState({});
 
-export function reducer(state = initialState, action: CampaignActions): State {
+export function reducer(state = initialState, action: CampaignActions | FlightPreviewActions): State {
   switch (action.type) {
     case ActionTypes.CAMPAIGN_NEW: {
       return { ...adapter.removeAll(state), campaignId: undefined };
@@ -145,6 +146,11 @@ export function reducer(state = initialState, action: CampaignActions): State {
         );
       }
       return newState;
+    }
+    case ActionTypes.CAMPAIGN_FLIGHT_PREVIEW_CREATE_SUCCESS: {
+      const { flight, status, statusMessage } = action.payload;
+      const localFlight = state.entities[flight.id].localFlight;
+      return adapter.updateOne({ id: flight.id, changes: { localFlight: { ...localFlight, status, statusMessage } } }, state);
     }
     default: {
       return state;
--- a/src/app/core/campaign/flight-preview.service.ts
+++ b/src/app/core/campaign/flight-preview.service.ts
@@ -1,54 +1,39 @@
 import { Injectable } from '@angular/core';
-import { Observable } from 'rxjs';
-import { map, mergeMap, first } from 'rxjs/operators';
+import { Observable, forkJoin, of } from 'rxjs';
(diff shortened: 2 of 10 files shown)


src/app/campaign/store/reducers/flight.reducer.ts
```diff
@@ -146,6 +147,11 @@ export function reducer(state = initialState, action: CampaignActions): State {
       }
       return newState;
     }
+    case ActionTypes.CAMPAIGN_FLIGHT_PREVIEW_CREATE_SUCCESS: {
+      const { flight, status, statusMessage } = action.payload;
```
cavis, line 151: I started out patching the entire previewed flight-doc back onto `localFlight`.  But that caused all kinds of crazy form flickering ... especially if you're changing fields fast and generating multiple in-progress preview POSTs.  Each time a 201 comes back, it resets your form back to that state!

So I went for the minimal status/message patching here.  Let me know what you think of that!

src/app/core/campaign/flight-preview.service.ts
```diff
@@ -1,54 +1,39 @@
...
         first(),
-        map(root => root.create('prx:flight-allocation-preview', {}, { ...flight, startAt, endAt, totalGoal, dailyMinimum, uncapped })),
-        mergeMap(flightPreviewDoc => flightPreviewDoc.followList('prx:flight-days'))
+        mergeMap(root => this.preview(flight, root, 'prx:flight-allocation-preview'))
       );
     }
   }
+
+  private preview(flight: {}, doc: HalDoc, link: string) {
+    return doc.create(link, {}, flight).pipe(
```
cavis, line 29: It seems fine to just pass the whole dang Flight model through to the haldoc `JSON.stringify()` in there.  Moment dates just work out.

src/app/campaign/store/reducers/flight.reducer.ts
sandikbarr: ugh, yeah I could see that being gross. But doesn't this also cause a state cycle update? Because even if you weren't updating the localFlight field, the whole state should be replaced on each update to the state, so do the selectors see this as a change? I'm not sure what the difference would be where you're seeing that flickering, seems like it would do the same thing with this.

I'm wondering if the status and statusMessage should be stored on the reducer entity FlightState instead of the flight model? I understand that you'd then have to pass it down to the inventory.component separate from the Flight Input, but previewError has its own selector and gets passed in there.

src/app/core/campaign/flight-preview.service.ts
sandikbarr: nice, that's great!

sandikbarr (commented):
I'm gonna take a look at the flickering

src/app/campaign/store/reducers/flight.reducer.ts
cavis: The difference is that the other `localFlight` fields are in the form, and status/message are not.  So with something like `totalGoal` -

1. Change `totalGoal = 10` in the form
2. Fires off POST preview
3. Change `totalGoal = 11` in the form
4. Fires off a 2nd POST preview
5. Preview (2) returns, changes form `totalGoal` back to 10.  _(I don't _think_ this fires off a 3rd POST preview - something is at least preventing that)_
6. Preview (4) returns, changes form `totalGoal` back to 11.

sandikbarr: I can see the flickering with that, ok. But hmm, you know it seems like what you describe in itself is a problem. Not your problem in this PR (actually my fault,) but we should be using `switchMap` in the `createFlightPreview$` @Effect instead of `mergeMap`. It should cancel the prior request with totalGoal of 10 and only respond with 11.

sandikbarr (approved):
This is good. I'll open a ticket for that switchMap/mergeMap.

src/app/campaign/store/reducers/flight.reducer.ts
cavis: Oh, cool!  I didn't even know that was a thing, but sounds like a good solution.  Thanks ngrx!

PR #204, opened 2020-05-28
190: Rearrange flight table columns

### LINKED ISSUE #190: Change order of flight table columns on dashboard
- C. Status
- F. Status
- Campaign Name
- Start
- End
- Actuals
- Goal
- Advertiser
- Show
- Zone
- Geotarget
- Type
- Rep

rpeterman-gp:
Closes #190 

Rearranges flight table columns to match the order in #190. Flight name was left as the first fixed column since removing it seemed inadvisable. (Campaign card list is already weird by not showing the campaign name on the card, right?)


sandikbarr (approved):
yes, totally. I think Brandon has taken that feedback about the campaign name not being shown, but we haven't done much of anything on those campaign cards since the initial designs. Seemed like getting this flight table right and thinking about other dashboard stuff was more the priority.

Anyway, thanks so much for picking this up!

PR #205, opened 2020-05-28
195: Hide delete button when flight has actuals

### LINKED ISSUE #195: Don't allow deletion of flights if there are actuals
Just like with campaign deletion, we shouldn't allow flights to be deleted if they have any actuals attached to them. Hide the delete button in this use case.

rpeterman-gp:
Closes #195 

This is a superficial solution. My assumption is the API should enforce this with an error response when trying to destroy a flight doc that has actuals, just in case someone manages to bypass this full-proof preventative measure.


sandikbarr (approved):
yep, totally. perfect, thanks!

PR #207, opened 2020-06-01
Contract goal/date fields

### LINKED ISSUE #105: Add Contract Goal field to flight form impressions table header
Add it to the impressions table header next to total goal field.

Inventory table header changes

cavis:
Closes #105.

Frontend for PRX/augury.prx.org#251 (and should merge after that).  I'm realizing I didn't exactly follow the design of that ticket - but the header has changed a bit since then?  Hoping that #208 can clean that up.

Adds 3 optional fields to flight form: "Contract Start", "Contract End", and "Contract Goal".

Capped flights: 
image

And uncapped flights (note I'm allowing a contract goal here even though you can't have an explicit total_goal ... #208 may clean this up): 
image

### DIFF
--- a/src/app/campaign/flight/flight-form.component.html
+++ b/src/app/campaign/flight/flight-form.component.html
@@ -6,14 +6,14 @@
     </mat-form-field>
     <div class="inline-fields">
       <mat-form-field appearance="outline">
-        <mat-label>Start Date</mat-label>
-        <input matInput [matDatepicker]="startPicker" [max]="flight?.endAt" placeholder="Start Date" formControlName="startAt" />
+        <mat-label>Actual Start Date</mat-label>
+        <input matInput [matDatepicker]="startPicker" [max]="flight?.endAt" placeholder="Actual Start Date" formControlName="startAt" />
         <mat-datepicker-toggle matSuffix [for]="startPicker"></mat-datepicker-toggle>
         <mat-datepicker #startPicker></mat-datepicker>
       </mat-form-field>
       <mat-form-field appearance="outline">
         <mat-label>Actual End Date</mat-label>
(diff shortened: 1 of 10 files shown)


src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -77,6 +69,22 @@ export class CampaignActionService implements OnDestroy {
     return startAt && endAt && startAt.valueOf() < endAt.valueOf();
   }
 
+  hasChanged(a: any, b: any) {
```
cavis, line 72: I was looking for a general solution to "check for form changes", and this is what I came up with.  Differences from a normal JSON.stringify comparison:

- "A" (the flight form) objects can be missing keys present in "B".  This is helpful for things like flight zones, where the form only has `[{id: "pre_1"}]` but the remote JSON might be `[{id: "pre_1", label: "Preroll 1", pingbacks: []}]`.
- Undefined and null are equal.  Since our HAL API tends to omit any null fields, but we end up PUT-ing back nulls from the form.

src/app/campaign/flight/flight-form.component.html
```diff
@@ -7,19 +7,43 @@
     <div class="inline-fields">
       <mat-form-field appearance="outline">
         <mat-label>Start Date</mat-label>
```
sandikbarr: Probably my fault here screwing up these labels (I think going back and further between the material datepickers and the styleguide prx-datepickers with the UTC support and everything,) but we should probably make the Actual Start and End date labels consistent.

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: This is an excellent improvement. Thanks for adding this.

sandikbarr (approved):

PR #210, opened 2020-06-05
angular 8 upgrade

### LINKED ISSUE #125: Upgrade to angular 8
Consider the right time to do this...

sandikbarr:
Closes #125


src/app/campaign/store/actions/account-action.creator.ts
```diff
@@ -1,21 +1,10 @@
-import { Action } from '@ngrx/store';
+import { createAction, props, union } from '@ngrx/store';
 import { ActionTypes } from './action.types';
```
sandikbarr, line 2: v8 NgRx Actions now use createAction to reduce boilerplate and improve type safety. https://medium.com/angular-in-depth/ngrx-action-creators-redesigned-d396960e46da
* action.payload is now obsolete, replaced by props which adds the properties onto the Action
* ActionTypes enum no longer required, in v9 createAction will check plain string values for the action types to be unique. https://twitter.com/tim_deschryver/status/1267849420898476032

src/app/campaign/store/actions/advertiser-action.creator.ts
```diff
@@ -1,43 +1,21 @@
...
-  | AddAdvertiserSuccess
-  | AddAdvertiserFailure;
+export const AdvertisersLoad = createAction(ActionTypes.CAMPAIGN_ADVERTISERS_LOAD);
+export const AdvertisersLoadSuccess = createAction(ActionTypes.CAMPAIGN_ADVERTISERS_LOAD_SUCCESS, props<{ docs: HalDoc[] }>());
+export const AdvertisersLoadFailure = createAction(ActionTypes.CAMPAIGN_ADVERTISERS_LOAD_FAILURE, props<{ error: any }>());
+
+export const AddAdvertiser = createAction(ActionTypes.CAMPAIGN_ADD_ADVERTISER, props<{ name: string }>());
+export const AddAdvertiserSuccess = createAction(ActionTypes.CAMPAIGN_ADD_ADVERTISER_SUCCESS, props<{ doc: HalDoc }>());
+export const AddAdvertiserFailure = createAction(ActionTypes.CAMPAIGN_ADD_ADVERTISER_FAILURE, props<{ error: any }>());
+
```
sandikbarr, line 12: The actions union is also no longer required for type safety when using effects and reducers with `createEffect ` and `createReducer` +  `on`. I used it for the transition and kept it because I have a hard time letting go.

src/app/campaign/store/actions/campaign-action.creator.ts
```diff
@@ -1,138 +1,90 @@
...
-export class CampaignAddFlight implements Action {
-  readonly type = ActionTypes.CAMPAIGN_ADD_FLIGHT;
-
-  public payload: { campaignId: number | string; flightId: number; startAt: Moment; endAt: Moment };
-
-  constructor(payload: { campaignId: number | string; flightId?: number; startAt?: Moment; endAt?: Moment }) {
+export const CampaignNew = createAction(ActionTypes.CAMPAIGN_NEW);
+
+export const CampaignDuplicate = createAction(ActionTypes.CAMPAIGN_DUPLICATE, props<{ campaign: Campaign; flights: Flight[] }>());
+
```
sandikbarr, line 10: createAction can take `props` or it can take a function that returns props as it's second parameter. props reduces boilerplate.

src/app/campaign/store/effects/account.effects.ts
```diff
@@ -1,21 +1,21 @@
...
 
 @Injectable()
 export class AccountEffects {
-  @Effect()
-  loadAccounts$ = this.actions$.pipe(
-    ofType(ActionTypes.CAMPAIGN_ACCOUNTS_LOAD),
-    switchMap(() => this.user.accounts),
-    withLatestFrom(this.user.defaultAccount),
-    map(([accounts, defaultAccount]) => new accountActions.AccountsLoadSuccess({ docs: [defaultAccount].concat(accounts) })),
-    catchError(error => of(new accountActions.AccountsLoadFailure({ error })))
```
sandikbarr, line 17: createEffect adds type safety. Fails with compile time error if the effect does not return an Action or have `{dispatch: false}` https://medium.com/ngrx/announcing-ngrx-version-8-ngrx-data-create-functions-runtime-checks-and-mock-selectors-a44fac112627

src/app/campaign/store/reducers/campaign.reducer.ts
```diff
@@ -21,115 +21,92 @@ export const initialState: CampaignState = {
...
-        localCampaign: {
-          ...state.localCampaign,
-          ...campaign
-        },
-        changed,
-        valid
-      };
-    }
-    case ActionTypes.CAMPAIGN_ADD_ADVERTISER_SUCCESS: {
-      const { set_advertiser_uri } = docToAdvertiser(action.payload.doc);
```
sandikbarr, line 62: `createReducer` uses `on` with action reducer functions instead of a switch
* it has built in type safety with action creators
* doesn't need a default case
* at some point, they also added support for processing an action type in more than one `on` so that if an action type is seen more than once, the reducer is wrapped and the additional `on` is applied. (not used in this app, just interesting and different than what you can do with switch)

src/app/campaign/store/actions/account-action.creator.ts
cavis: +10 nice.  take that, boilerplate!  (Also a good article - nice to see actual explanations of API changes)

src/app/campaign/store/effects/campaign.effects.ts
```diff
@@ -1,277 +1,287 @@
...
+            ({
+              campaignDoc,
+              deletedFlightDocs,
+              createdFlightDocs
+            }: {
+              campaignDoc: HalDoc;
+              deletedFlightDocs: { [id: number]: HalDoc };
+              createdFlightDocs: { [id: number]: HalDoc };
+            }) => {
+              this.toastr.success('Campaign saved');
```
cavis, line 155: This mega-save effect makes a fun diff when the indentation changes slightly.

But just viewing the file... yeah, looks pretty much the same to me!

src/app/campaign/store/models/flight.models.ts
```diff
@@ -52,15 +52,19 @@ export const getFlightZoneIds = (zones: any[]): string[] => {
   return zones.filter(z => z).map(z => z.id || z);
 };
 
-export const duplicateFlight = (flight: Flight, tempId: number): Flight => {
+export const duplicateFlightState = (flight: Flight, tempId: number, changed, valid): FlightState => {
```
cavis, line 55: Ooh, also catching the previous form state when copying to a new form?  I'm guessing the new form looks pristine when you create it ... good catch!

src/app/campaign/store/reducers/account.reducer.ts
```diff
@@ -9,22 +9,16 @@ export interface State extends EntityState<Account> {
...
-    }
-  }
+// tslint:disable-next-line: variable-name
+const _reducer = createReducer(
+  initialState,
+  on(accountActions.AccountsLoadSuccess, (state, action) => ({ ...adapter.addAll(action.docs.map(docToAccount), state), error: null })),
+  on(accountActions.AccountsLoadFailure, (state, action) => ({ ...state, error: action.error }))
+);
+export function reducer(state, action) {
+  return _reducer(state, action);
```
cavis, line 21: This bit is kind of funny looking ... is this a common pattern?  I'm not sure I follow how this is any different than just an `export const reducer = createReducer(...)`

cavis (approved):
:+100: one question about the `_reducer` pattern, but otherwise all looks good.  And thanks for the article links!

src/app/campaign/store/reducers/account.reducer.ts
sandikbarr: It is funny looking, and I don't like it. But this was the problem with the AOT build. https://medium.com/angular-in-depth/solving-aot-error-in-ngrx-function-calls-are-not-supported-in-decorators-5c337381457a

Apparently this goes away with Ivy in ng9

Issue #133, opened 2020-02-18
Flight episode targeting


cavis:
Target specific episodes from a flight.  After #132.  And maybe after or concurrently with #131 since they share a lot in common.

Somewhere in here:

![image](https://user-images.githubusercontent.com/1410587/74776640-2bdfb380-5255-11ea-8609-5da70c5616d7.png)

Will look very similar to the "geo-targeting" UI component.

- [ ] Include/exclude option
- [ ] Search box for an episode guid (#132).

Maybe we should consolidate Locations, Episodes, and Keywords into a generic "Targeting" section, instead of those checkboxes?  For every target, you select include/exclude, a type (Country, Episode, Metro, Keyword etc), and a value.

Or maybe that's bad UI.  @brandonhundt ?

PR #213, opened 2020-06-18
Flight targets

### LINKED ISSUE #131: Flight location targeting
Requires PRX/augury.prx.org#176.  Implement "location targeting UI".  Including:

- Include/exclude boolean option
- A "type" option, which is Country/Subdiv/Metrocode.
- More dropdown(s) for the code.
- TODO: where to get the type/code options?  Hardcode in Grove for now, probably.
- Each row can have only 1 "type" and 1 "code"

Something like, but not exactly like this:

image

cavis:
Closes #131 and #133.  Adds optional has-many "Targets" to a Flight.  This implements the "Episode" and "Countries" currently in Augury, but others should be easy to add.

image

- [x] Convert "inventory" options loading to ngrx
- [x] Add "inventory targets" options loading to ngrx
  - Each inventory record has a `prx:targets` link, which is loaded on demand when you navigate to a flight against that inventory.
- [x] UI for adding/removing targets
  - I attempted to do an autocomplete/searchable field for the "code" (Country name or Episode title) ... but I couldn't make it work very well.  Maybe we should take this as new ticket out of this?  For now, I just sort the options.
- [x] Add a "loading" spinner to the inventory table
  - Turns out, previewing "exclude" targets can be slow, and it's a bit hard to see what's going on without a spinner.

### DIFF
--- a/src/app/campaign/flight/flight-targets-form.component.ts
+++ b/src/app/campaign/flight/flight-targets-form.component.ts
@@ -42,6 +42,7 @@ import moment from 'moment';
   ]
 })
 export class FlightTargetsFormComponent implements ControlValueAccessor {
+  // tslint:disable-next-line
   private _targetOptions: InventoryTargets;
   get targetOptions() {
     return this._targetOptions;
@@ -86,6 +87,7 @@ export class FlightTargetsFormComponent implements ControlValueAccessor {
     this.onTouchedFn = fn;
   }
 
+  // tslint:disable-next-line
   validate(_control: FormControl) {
     return !this.targets || this.targets.valid ? null : { error: 'Invalid targets' };
   }
@@ -125,10 +127,10 @@ export class FlightTargetsFormComponent implements ControlValueAccessor {
 
   private filteredTargetOptions(type: string, code: string, currentTargets: FlightTarget[]) {
     return this.targetCodesForType(type).filter(opt => {
-      if (opt.type == type && opt.code == code) {
+      if (opt.type === type && opt.code === code) {
         return true;
       } else {
-        return !currentTargets.find(t => opt.type == t.type && opt.code == t.code);
+        return !currentTargets.find(t => opt.type === t.type && opt.code === t.code);
       }
     });
   }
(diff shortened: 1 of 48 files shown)


src/app/campaign/flight/flight-targets-form.component.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  private setCodeOptions(targets: FlightTarget[]) {
+    this.codeOptions = (targets || []).map(target => {
+      return this.filteredTargetOptions(target.type, target.code, targets);
+    });
+  }
+
+  private filteredTargetOptions(type: string, code: string, currentTargets: FlightTarget[]) {
+    return this.targetCodesForType(type).filter(opt => {
+      if (opt.type == type && opt.code == code) {
```
sandikbarr: what is this `==` business? ;) linter should catch that, but only if your editor is using it

src/app/core/inventory/inventory.service.ts
```diff
@@ -1,81 +1,21 @@
...
-        return inventory.follow('prx:availability', {
-          startDate,
-          endDate,
-          zone,
-          flightId
-        });
-      })
-    );
+  loadInventoryTargets(inventory: number | HalDoc): Observable<HalDoc> {
+    if (inventory instanceof HalDoc) {
```
sandikbarr: I'm sure this is totally fine, but there's a lot of distrust for `instanceof` in JS. It's good for type guards in TS, but I try to avoid it for runtime checks.

sandikbarr (approved):
looks great, really nice work on this

src/app/campaign/flight/flight-targets-form.component.ts
cavis: Hrmm.... yeah, what the heck?  I'll check that my linter is actually fixing these files.

src/app/core/inventory/inventory.service.ts
cavis: Interesting - maybe I'll add a 2nd runtime-ish check, just in case!

PR #215, opened 2020-06-23
flight preview switchMap

### LINKED ISSUE #203: On new preview, flight preview should cancel any other in progress flight preview API requests
Seems like the `createFlightPreview$` Effect should be using `switchMap` instead of `mergeMap` to cancel and replace any in progress API requests.

1. Change totalGoal = 10 in the form
2. Fires off POST preview
3. Change totalGoal = 11 in the form
4. Fires off a 2nd POST preview
5. Preview (2) returns, changes form totalGoal back to 10. (I don't think this fires off a 3rd POST preview - something is at least preventing that)
6. Preview (4) returns, changes form totalGoal back to 11.

sandikbarr:
Closes #203 

If you kick off more than one flight preview post, you'll see the first ones get cancelled and replaced by subsequent posts if they are still ongoing. Previously this was causing those responses to get intermingled, but we only want the last one. (If we wanted the first one and to ignore any subsequent posts, the operator to use would be exaustMap.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@ngrx/store-devtools": "~8.6.0",
     "angulartics2": "^7.5.2",
     "ngx-prx-styleguide": "3.0.0",
-    "prx-ng-serve": "^3.0.1",
+    "prx-ng-serve": "^3.0.2",
     "rxjs": "~6.5.5",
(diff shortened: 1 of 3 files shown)


cavis (approved):
That's all it takes?  Neat!  👍

Issue #199, opened 2020-05-19
Untouched new flight looks valid


cavis:
Steps:

1) new campaign
2) new flight
3) back to campaign form - fill out
4) click save
5) “Campaign Forbidden” error message, but you stay on the `/campaign/new` page.

Expected:

- [ ] Dismissible error message thing for error on save
- [ ] Still attempt to redirect to the new campaign if a flight errors (it's much more common for us to see flight errors than campaign - and it's easy to lose track of the campaign if it has no flights attached)
- [x] Make sure new/untouched flight forms look invalid by default somehow.

sandikbarr:
error handling moved to #222

PR #217, opened 2020-06-30
fixes untouched flight looks valid

sandikbarr:
addresses one part of #199 and fixes a slight bug with loading in flight preview error/failure

* I have a branch that does the dismissible error on save, but I'm struggling with unit tests on the Material Snackbar.
* Also in this branch, I am putting flight zones into a separate control value accessor component, so far a disaster and I'm reading and re0reading the articles and trying all the things. The current flight zones implementation still feels off to me and has some issues with not following unidirectional data flow which creates multiple update cycles, same with targets.

### DIFF
--- a/src/app/campaign/store/reducers/flight.reducer.ts
+++ b/src/app/campaign/store/reducers/flight.reducer.ts
@@ -59,7 +59,7 @@ const _reducer = createReducer(
         set_inventory_uri: null
       },
       changed: false,
-      valid: true
+      valid: false
     };
     return adapter.addOne(initialFlightState, state);
   }),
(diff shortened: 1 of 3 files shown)


src/app/campaign/store/reducers/flight.reducer.ts
```diff
@@ -59,7 +59,7 @@ const _reducer = createReducer(
         set_inventory_uri: null
       },
       changed: false,
```
sandikbarr, line 61: This right here was the new flight bug. Ryan pointed a few of these out in PRs earlier in the project, think I wouldn've gotten that sorted by now.

cavis (approved):
👍

PR #218, opened 2020-06-30
211: Add companion toggle to zones

### LINKED ISSUE #211: companion toggle
add support to set the companion toggle on zone

rpeterman-gp:
Closes #211 

### This PR...
- Adds slide toggle input for `isCompanion` prop to flight form.
- Adds `isCompanion` to Flight model.
- Adds `isCompanion` to the props that trigger an allocation preview.
- Updates flight-form spec.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,9 @@
 {
-    "editor.formatOnSave": true
+  "editor.formatOnSave": true,
+  "prettier.packageManager": "yarn",
+  "prettier.prettierPath": "./node_modules/prettier",
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "[html]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  }
 }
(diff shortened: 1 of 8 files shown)


sandikbarr (changes requested):
Can you make sure you're using the .prettier configuration in the project? I keep seeing these formatting changes especially to the templates in your PRs. I'm working on this same code, and it really makes merge conflicts more difficult.

rpeterman-gp:
I am setup to use the prettier config in the project. The template HTML files would actually not be formatted by prettier, but by the editor's HTML settings, which I do not think we are capturing in the committed .vscode settings. Looks like the unintended changes caused by my HTML formatting settings is the `html.format.wrapAttributes` setting. Mine is set at the user level to the default of `auto`, but it looks like that file had previously been formatted with `aligned-multiple`. Can we confirm that is the setting someone who has touch these files has been using, and if that is the setting that should be used. I can add it to the .vscode settings in the project.

sandikbarr (approved):
Looks great! Thanks so much for sorting out that formatting, will be very helpful for comparing and merging this into my WIP zones ControlValueAccessor component.

PR #219, opened 2020-07-01
fudge end date

### LINKED ISSUE #175: fudge flight end date to be inclusive
The user expects the end date to be included in the forecast, but the end date is Augury's cut off, exclusive of that date. The front end should display the end date as one day before the API date and save to the API as the actual exclusive end date.

sandikbarr:
Closes #175 

The cutoff is at midnight, which makes the flight stop serving a day _before_ the end date. This is confusing to folks and could throw off reporting. "Fudges" the flight end date by one day to make it appear as though the end date is the same day the flight stops serving.


src/app/campaign/flight/flight-form-control-container.component.ts
```diff
@@ -89,7 +89,7 @@ export class FlightFormControlContainerComponent implements OnInit, OnDestroy {
     id: [],
     name: ['', Validators.required],
     startAt: ['', Validators.required],
-    endAt: ['', Validators.required],
+    endAtFudged: ['', Validators.required],
     contractStartAt: [''],
     contractEndAt: [''],
```
cavis, line 94: Hrmm... I wonder if we should do the same thing with the contract end date.  They'll be setting it using the same "inclusive end date" logic.

src/app/campaign/store/models/flight.models.ts
```diff
@@ -50,6 +51,8 @@ export const docToFlight = (doc: HalDoc): Flight => {
   const flight = filterUnderscores(doc) as Flight;
   flight.startAt = utc(flight.startAt);
   flight.endAt = utc(flight.endAt);
+  // the cutoff is midnight the following day, but want it to appear as if the flight ends on the date it is set to stop serving
+  flight.endAtFudged = utc(flight.endAt).subtract(1, 'days');
```
cavis: Okay, so if i'm following this ...

1. This is a one-time transform of endAt -> endAtFudged when the remote HalDoc comes in.
2. Form field is bound to endAtFudged.
3. On form change, the CampaignActionService transforms endAtFudged back into endAt, which gets sent on any remote requests.  And we don't get an infinite loop since (1) only happens after a remote load.

cavis (changes requested):
+1 succinct!  And naming makes it obvious what craziness is happening.

Should probably also do `contract_end_at` in this PR, which will hopefully be simpler (fewer tests) ... but then I think this can merge.

src/app/campaign/store/models/flight.models.ts
sandikbarr: Yes! I had kind of thought at first we would hide it in the service layer, but I did that and it felt gross. And the dashboard isn't using the same service layer anyway.
(1) will happen on load and on flight form update, but yeah those updates should not be emitting any further updates.

I am still seeing a separate bug though where we do emit a flight form update on flight tab navigation, which causes a flight preview load.

cavis (approved):
:+100:

PR #223, opened 2020-07-07
flight zones ControlValueAccessor component

sandikbarr:
Separates flight zones into a ControlValueAccessor component. Uses unidirectional data flow to add and remove zones, to update the form, or when navigating to different flight tabs. Zone changes get dispatched out to the store, and those updates come back in through a patchValue on the parent form component.


src/app/campaign/flight/flight-form-control-container.component.ts
```diff
@@ -128,42 +119,8 @@ export class FlightFormControlContainerComponent implements OnInit, OnDestroy {
...
-    return this.flightForm.get('zones') as FormArray;
-  }
-
-  createZoneControl() {
-    return this.fb.group({ id: ['', Validators.required], url: ['', validateMp3] });
-  }
-
-  onAddZone() {
-    // create new form group/zone control with initial empty values
-    const newZone = this.createZoneControl();
```
sandikbarr, line 141: This was causing two update cycles. One with the empty zone, and second with the control reset to the first option. This is one of the things that led me down the unidirectional data flow way of thinking. Maybe the intention was not to cause an update cycle at all by manipulating the form?

src/app/campaign/flight/flight-targets-form.component.ts
```diff
@@ -73,6 +73,8 @@ export class FlightTargetsFormComponent implements ControlValueAccessor {
   writeValue(targets: FlightTarget[]) {
     this.setCodeOptions(targets);
     this.targets = new FormArray((targets || []).map(this.newFlightTarget));
+    // TODO: wouldn't this be leaking subscriptions? Instead of new'ing the FormArray and subscribing on each update,
+    //  it should update/add/remove the controls in the array
```
sandikbarr, line 77: This was another thing that concerned me with how we were subscribing each time the form was updated.

rpeterman-gp (approved):
Looks great! Very educational review. Can't wait to apply this pattern to the targets component.

PR #224, opened 2020-07-09
validates flight dates against actuals served

### LINKED ISSUE #75: flight start/end date validation
- [x] Flight dates should be validated so that end at must be greater than start at. User should not be able to Save the form unless valid.
- [ ] If flight dates are changed, validate that the start and end date fall within the range of actuals that have been served
- (are there any other validation rules like not setting up new flights in the past or editing flights to be further in the past than has already run?)

sandikbarr:
Closes #75 

Validates the flight start and end dates to be within the dates that actuals have been served. It doesn't stop the preview request on flight date change, which also gives the same error, but it does prevent the flight from being saved because of an invalid form.

Flight form fields were getting touched reset on form updates because of the use of `reset` in the incoming form updates, so the validation errors would flash up but then go away. I changed that `reset` to `patchValue` to resolve this. Maybe when the flight id changes, the form controls could be reset when updating the values to the routed flight if it hasn't been changed, but probably the correct way to have done this would have been to have all the flight forms in a single parent form group instead of swapping out the one form's values.

### DIFF
--- a/src/app/campaign/flight/flight-form-control-container.component.ts
+++ b/src/app/campaign/flight/flight-form-control-container.component.ts
@@ -1,7 +1,8 @@
 import { Component, OnInit, OnDestroy, Input, Output, EventEmitter, ChangeDetectionStrategy } from '@angular/core';
-import { FormBuilder, Validators } from '@angular/forms';
+import { AbstractControl, FormBuilder, Validators } from '@angular/forms';
 import { Subscription } from 'rxjs';
 import { Flight, FlightZone, InventoryRollup, Inventory, InventoryZone, InventoryTargets } from '../store/models';
+import { utc } from 'moment';
 
 @Component({
   selector: 'grove-flight',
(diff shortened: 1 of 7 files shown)


src/app/campaign/flight/flight-form-control-container.component.ts
```diff
@@ -121,7 +146,7 @@ export class FlightFormControlContainerComponent implements OnInit, OnDestroy {
 
   // updates the form from @Input() set flight
   setFlightForm(flight: Flight) {
-    // reset the form
```
sandikbarr, line 124: if we use `reset` it also resets the controls to pristine, which doesn't seem like something we want on every update cycle

src/app/campaign/store/selectors/flight-days.selectors.ts
```diff
@@ -131,3 +131,18 @@ export const selectFlightDaysRollup = createSelector(
...
+  const startAt =
+    flightDays &&
+    flightDays.days &&
+    [...flightDays.days].sort((a, b) => a.date.valueOf() - b.date.valueOf()).find(day => day.numbers.actuals > 0);
+  const endAt =
+    flightDays &&
+    flightDays.days &&
+    [...flightDays.days].sort((a, b) => b.date.valueOf() - a.date.valueOf()).find(day => day.numbers.actuals > 0);
+  if (startAt && endAt) {
+    console.log({ startAt: startAt.date, endAt: endAt.date });
```
rpeterman-gp: Do we need this console.log?

rpeterman-gp (changes requested):
Looks good. Only thing I found was a console.log statement we probably don't need in prod.

src/app/campaign/store/selectors/flight-days.selectors.ts
sandikbarr: oops dang, will remove!!

rpeterman-gp (approved):

PR #232, opened 2020-07-20
Fixes preview loading on flight tab switch

sandikbarr:


src/app/campaign/flight/flight-zones-form.component.ts
```diff
@@ -129,16 +129,13 @@ export class FlightZonesFormComponent implements ControlValueAccessor, OnDestroy
   }
 
   onAddZone() {
-    // this will send an Add Zone action dispatch up thru the state to be added to the flight,
-    //  and the zone control will be added to the form in writeValue
-    // add the zone that's at the top of the options list
-    this.addZone.emit({ zone: filterZones(this.zoneOptions, this.flightZones as InventoryZone[]).shift() });
```
sandikbarr, line 135: further discussion with @rpeterman-gp determined this is how we're gonna do form arrays, this results in CAMPAIGN_FLIGHT_FORM_UPDATE and CAMPAIGN_FLIGHT_PREVIEW_CREATE because the update will go out through the parent form's valueChanges stream (but only once per update ;)

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -93,6 +93,8 @@ export class CampaignActionService implements OnDestroy {
         (a === undefined || a === null || a === '' || a.length === 0) &&
         (b === undefined || b === null || b === '' || b.length === 0)
       );
+    } else if (isMoment(a) && isMoment(b)) {
```
sandikbarr, line 96: comparing the endAt values as Object was always giving false positive for hasChanged

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -117,12 +132,15 @@ export class CampaignActionService implements OnDestroy {
...
 
   havePreviewParamsChanged(a: Flight, b: Flight) {
-    const check = ['startAt', 'endAt', 'set_inventory_uri', 'zones', 'targets', 'totalGoal', 'dailyMinimum', 'deliveryMode', 'isCompanion'];
-    return check.some(fld => {
-      if (this.hasChanged(b[fld], a[fld])) {
-        return true;
-      }
-    });
+    const check = ['startAt', 'endAt', 'set_inventory_uri', 'targets', 'totalGoal', 'dailyMinimum', 'deliveryMode', 'isCompanion'];
+    return (
```
sandikbarr, line 136: Sometimes the flight zones from the API has pingbacks that the form doesn't know about. The only thing that requires a new preview is add/remove/update a zone id, not urls, not pingbacks. Where we are coercing the InventoryZone into a FlightZone is perhaps also troublesome. I saw where some had labels and others didn't.

rpeterman-gp: It would be nice to at some point convert this logic into a pipe-like execution of validation functions that can handle complex logic for comparing changes to values like zones and creatives, which may not need preview updates for every items' value changes.

src/app/campaign/store/actions/campaign-action.service.ts
rpeterman-gp: Nice! I had noticed that I was always getting an extra change cycle after the first change of any field due to the moment objects differing.

sandikbarr: yeah, saw a couple things once I stepped thru it a little bit and that was one of em, comparing the whole Moment and being off 😬

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: yeah I was also thinking that there should be separate streams for form updates and loading the preview because the logic and model is just getting more and more complex

cavis (approved):
👍  one issue with null total_goals ... but otherwise looks good to me!

sandikbarr:
thanks for the PR, gonna merge this and get it into my pingbacks branch that's in progress

sandikbarr:
added #233 for the total goal issue

PR #237, opened 2020-07-21
Dashes for past availability

### LINKED ISSUE #221: Past availability should be dashes
Related to PRX/augury.prx.org#277.

As a quick patch, let's just put in dashes instead of zeroes for past availability.

cavis:
Fixes #221.

Also adds the largeNumber pipe to the total-actuals in the lower right of the table.


sandikbarr (approved):
looks good!

PR #238, opened 2020-07-22
form array guard and other fixes

### LINKED ISSUE #228: `selectChange` selector not using current flights' changes status
### Reproduce

1. Edit a flight on a campaign with multiple flights.
2. Make a change to the title of the flight.
3. Note the save button is now active. DO NOT PRESS THE BUTTON.
4. Click on another flight tab.
5. Note the save button is now disabled.
6 Switch back to the edited flight.
7. Note save button is still disabled, though changes are still properly shown in form field and on tab label.

#### Expected Behavior
Once a valid change has been made to a flight, the save button for the campaign should remain active even when navigated to another unmodified flight.

### LINKED ISSUE #233: total goal carried to flights with no total goal
If a campaign with multiple flights has a flight with no total goal, and you navigate between flight tabs, the total goal can get carried over from other flights. Able to recreate on staging with campaign/1803/flight/2108.

When converting from HalDoc, fields on the model that can be empty/null/undefined/not present should be explicitly set on the Flight. Those properties will not be present on the HalDoc because our Hal APIs don't render nil properties.

sandikbarr:
* closes #233 ensures nullable fields on flight are set in order to support form re-use
* closes #228 
  * adds an emitGuard to form array components for wrapping incoming updates
  * adds a filter to the flight form update stream to check that form flight and routed flight ids match
* uses an errorStateMatcher with the zones url (wasn't showing up as invalid)
* ensures that the creative url is cleared for empty zones
* uses onlySelf in ControlValueAccessor to update self but not parent
* removes the Flight Not Found error that started popping up frequently when the check for form save errors moved into the error service with the dismissible snackbar. This error intention was for an invalid flight id route, giving a lot of false positives, so maybe we can readdress at some point.

### DIFF
--- a/src/app/campaign/flight/flight-zones-form.component.ts
+++ b/src/app/campaign/flight/flight-zones-form.component.ts
@@ -120,7 +120,7 @@ export class FlightZonesFormComponent implements ControlValueAccessor, OnDestroy
       this.zones.markAsPristine();
     }
     // patch all of the values onto the FormArray
-    this.zones.patchValue(zones || [], { emitEvent: false });
+    this.zones.patchValue(zones, { emitEvent: false, onlySelf: true });
     this.emitGuard = false;
   }
 
--- a/src/app/campaign/flight/flight-zones-form.component.ts
+++ b/src/app/campaign/flight/flight-zones-form.component.ts
@@ -108,13 +108,13 @@ export class FlightZonesFormComponent implements ControlValueAccessor, OnDestroy
     // don't emit while manipulating FormArray with incoming update
     this.emitGuard = true;
     // get the correct number of zone fields and patchValue
-    const expectedLength = zones.length || 1;
-    while (this.zones.controls.length > expectedLength) {
+    zones = zones && zones.length ? zones : [{ id: '', url: '' }];
+    while (this.zones.controls.length > zones.length) {
       this.zones.removeAt(this.zones.controls.length - 1);
       this.zones.markAsPristine();
     }
     const addZones = [...zones];
-    while (this.zones.controls.length < expectedLength) {
+    while (this.zones.controls.length < zones.length) {
       // this only adds controls, doesnt reset controls already in form
       this.zones.push(this.flightZoneFormGroup(zones && addZones.pop()));
       this.zones.markAsPristine();
--- a/src/app/campaign/flight/flight-zones-form.component.ts
+++ b/src/app/campaign/flight/flight-zones-form.component.ts
@@ -45,7 +45,13 @@ export const validateMp3 = (value: string): { [key: string]: any } | null => {
         </mat-form-field>
         <mat-form-field appearance="outline">
           <mat-label>Creative Url</mat-label>
-          <input matInput formControlName="url" autocomplete="off" placeholder="Leave blank for a 0-second MP3" />
+          <input
+            matInput
+            [errorStateMatcher]="matcher"
+            formControlName="url"
+            autocomplete="off"
+            placeholder="Leave blank for a 0-second MP3"
+          />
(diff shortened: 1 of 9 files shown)


src/app/campaign/flight/flight-zones-form.component.ts
```diff
@@ -96,20 +105,23 @@ export class FlightZonesFormComponent implements ControlValueAccessor, OnDestroy
...
     const addZones = [...zones];
-    while (this.zones.controls.length < expectedLength) {
+    while (this.zones.controls.length < zones.length) {
       // this only adds controls, doesnt reset controls already in form
       this.zones.push(this.flightZoneFormGroup(zones && addZones.pop()));
       this.zones.markAsPristine();
     }
     // patch all of the values onto the FormArray
-    this.zones.patchValue(zones || [], { emitEvent: false });
+    this.zones.patchValue(zones, { emitEvent: false, onlySelf: true });
```
cavis, line 123: So this prevents triggering a form-changed event on the parent form?

src/app/campaign/flight/flight-zones-form.component.ts
```diff
@@ -45,7 +45,13 @@ export const validateMp3 = (value: string): { [key: string]: any } | null => {
         </mat-form-field>
         <mat-form-field appearance="outline">
           <mat-label>Creative Url</mat-label>
-          <input matInput formControlName="url" autocomplete="off" placeholder="Leave blank for a 0-second MP3" />
+          <input
+            matInput
+            [errorStateMatcher]="matcher"
```
cavis, line 50: So what's the difference between this and the default behavior?  As far as i can tell, this makes the error message appear right away rather than on-blur.  Just making sure I'm on the right track here.

(And I remember angrily trying to get that error message to show before-blur when working on the zones input ... good to know there's a way to fix that).

src/app/campaign/store/actions/campaign-action.service.ts
```diff
@@ -31,6 +31,8 @@ export class CampaignActionService implements OnDestroy {
     map(this.transformFlightForm),
     // if preview params changed, dispatch loadFlightPreview
     withLatestFrom(this.store.pipe(select(selectRoutedFlight))),
+    // re-using the flight form, so check that ids match
+    filter(([formState, flightState]) => !flightState || !flightState.localFlight || formState.flight.id === flightState.localFlight.id),
```
cavis, line 35: I'm not sure I ever noticed this error... but seems like a good sanity check! +2.

cavis (approved):
+103.  Couple questions, but all looks good to me!  Tested locally a bit, and things do seem more stable than staging.

src/app/campaign/flight/flight-zones-form.component.ts
sandikbarr: yeah, I guess that's what onlySelf is for, just update this control and any children. I think in practice, it's not doing anything extra for us here with the emit guard, but best practices and all (once you know better.)

src/app/campaign/store/actions/campaign-action.service.ts
sandikbarr: I only caught it a few times, and it would get overwritten by a subsequent update, but it seemed like a good one to catch for sure!

src/app/campaign/flight/flight-zones-form.component.ts
sandikbarr: The default behavior is to show errors when invalid and dirty. I was seeing invalid values in the field that weren't highlighted, and I couldn't see right away why the form was invalid. (I think the value may have carried over from another flight with that carry over bug, but it bothered the hecc out of me so I made it show the validation error.)

PR #244, opened 2020-07-24
fixes creative url carryover

### LINKED ISSUE #243: creative url carries over from other flights when empty
One a campaign with more than one flight where at least one flight has an empty creative field and another has a creative URL set, the creative URL will carry over into a flight with an empty creative URL from another flight when navigating between flights. This is related to the HAL API not including nullable fields.

sandikbarr:
closes #243 

Navigate between tabs on a campaign with multiple flights where some zones have creative urls and some don't. See that the creative urls stay with their own flight and don't get copied into new or other flights when navigating.

### DIFF
--- a/src/app/campaign/store/models/flight.models.ts
+++ b/src/app/campaign/store/models/flight.models.ts
@@ -65,7 +65,8 @@ export const docToFlight = (doc: HalDoc): Flight => {
     // the Flight model needs to have these fields in order to set up the flight form that is re-used between flights
     totalGoal: flight.hasOwnProperty('totalGoal') ? flight.totalGoal : null,
     dailyMinimum: flight.hasOwnProperty('dailyMinimum') ? flight.dailyMinimum : null,
-    contractGoal: flight.hasOwnProperty('contractGoal') ? flight.contractGoal : null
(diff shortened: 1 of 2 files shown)


src/app/campaign/store/models/flight.models.ts
```diff
@@ -65,7 +65,8 @@ export const docToFlight = (doc: HalDoc): Flight => {
     // the Flight model needs to have these fields in order to set up the flight form that is re-used between flights
     totalGoal: flight.hasOwnProperty('totalGoal') ? flight.totalGoal : null,
     dailyMinimum: flight.hasOwnProperty('dailyMinimum') ? flight.dailyMinimum : null,
-    contractGoal: flight.hasOwnProperty('contractGoal') ? flight.contractGoal : null
+    contractGoal: flight.hasOwnProperty('contractGoal') ? flight.contractGoal : null,
+    zones: flight.zones.map(({ id, label, url, fileSize, mimeType }) => ({ id, label, url, fileSize, mimeType }))
```
sandikbarr, line 69: mapping to ensure that zones has all its properties

cavis (approved):
👍 nice fix!

PR #245, opened 2020-07-24
flight zone pingbacks

### LINKED ISSUE #150: Implement pingbacks field
Implement design of #146.

Needs API changes in PRX/augury.prx.org#199.

- [ ] Ability to add/remove pingbacks for a flight zone
- [ ] Preview filled-in pingback url templates
- [ ] Validate url-templates

sandikbarr:
closes #150

### DIFF
--- a/src/app/campaign/flight/flight-form.component.html
+++ b/src/app/campaign/flight/flight-form.component.html
@@ -71,7 +71,6 @@
         [zoneOptions]="zoneOptions"
         [campaignId]="campaignId"
         [flightId]="flight?.id"
-        [podcastId]="podcastId"
       ></grove-flight-zones>
     </div>
     <div class="mat-form-field-wrapper flight-targets">
--- a/src/app/campaign/flight/flight-form.component.html
+++ b/src/app/campaign/flight/flight-form.component.html
@@ -69,6 +69,9 @@
         [flightZones]="flight?.zones"
         [isCompanion]="flight?.isCompanion"
         [zoneOptions]="zoneOptions"
+        [campaignId]="campaignId"
+        [flightId]="flight?.id"
+        [podcastId]="podcastId"
       ></grove-flight-zones>
     </div>
     <div class="mat-form-field-wrapper flight-targets">
--- a/src/app/campaign/flight/flight-form.component.ts
+++ b/src/app/campaign/flight/flight-form.component.ts
@@ -57,8 +57,4 @@ export class FlightFormComponent implements OnInit {
   checkError(fieldName: string, type = 'error') {
     return this.flightForm.get(fieldName).getError(type);
   }
-
-  get podcastId() {
-    return this.flight && this.flight.set_inventory_uri && this.flight.set_inventory_uri.split('/').pop();
-  }
 }
--- a/src/app/campaign/flight/flight-form.component.ts
+++ b/src/app/campaign/flight/flight-form.component.ts
@@ -30,6 +30,7 @@ export class FlightFormComponent implements OnInit {
   @Input() targetTypes: InventoryTargetType[];
   @Input() targetOptionsMap: InventoryTargetsMap;
   @Input() softDeleted: boolean;
+  @Input() campaignId: number;
   @Output() flightDuplicate = new EventEmitter<Flight>(true);
   @Output() flightDeleteToggle = new EventEmitter(true);
   flightForm: FormGroup;
@@ -55,5 +56,9 @@ export class FlightFormComponent implements OnInit {
 
   checkError(fieldName: string, type = 'error') {
     return this.flightForm.get(fieldName).getError(type);
+  }
+
+  get podcastId() {
+    return this.flight && this.flight.set_inventory_uri && this.flight.set_inventory_uri.split('/').pop();
   }
 }
--- a/src/app/campaign/flight/flight-form.error-state-matcher.ts
+++ b/src/app/campaign/flight/flight-form.error-state-matcher.ts
@@ -1,0 +1,10 @@
+import { ErrorStateMatcher } from '@angular/material';
+import { FormGroupDirective, FormControl } from '@angular/forms';
+
+export class FlightFormErrorStateMatcher implements ErrorStateMatcher {
+  isErrorState(control: FormControl | null, form: FormGroupDirective): boolean {
+    // form controls are showing as dirty but not as touched until after blur which isn't great for date pickers and validate on change
+    // so this ErrorStateMatcher overrides material to show errors when these controls are invalid and dirty (or touched)
+    return control && control.invalid && (control.dirty || control.touched);
+  }
+}
--- a/src/app/campaign/flight/flight-zones-form.component.scss
+++ b/src/app/campaign/flight/flight-zones-form.component.scss
@@ -5,11 +5,9 @@
   display: grid;
   grid-template-columns: 1fr 2fr 1fr 40px;
   grid-column-gap: 0.5rem;
+  align-items: center;
 }
 .remove-zone {
-  display: flex;
-  align-items: center;
-  justify-content: center;
   button {
     color: mat-color($prx-gray, 300);
     &:hover {
--- a/src/app/campaign/flight/flight-zones-form.component.spec.ts
+++ b/src/app/campaign/flight/flight-zones-form.component.spec.ts
@@ -4,8 +4,10 @@ import { FormBuilder } from '@angular/forms';
 import { MatFormFieldModule, MatSelectModule, MatInputModule, MatButtonModule, MatIconModule } from '@angular/material';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { ReactiveFormsModule } from '@angular/forms';
+import { PingbackFormComponent } from './pingbacks/pingback-form.component';
+import { FlightZonePingbacksFormComponent } from './pingbacks/flight-zone-pingbacks-form.component';
 import { FlightZonesFormComponent } from './flight-zones-form.component';
-import { FlightZone, InventoryZone, filterZones } from '../store/models';
+import { FlightZone, InventoryZone } from '../store/models';
 
 @Component({
   template: `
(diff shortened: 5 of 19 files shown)


src/app/campaign/flight/flight-form.component.html
```diff
@@ -69,6 +69,9 @@ <h2>Zones</h2>
         [flightZones]="flight?.zones"
         [isCompanion]="flight?.isCompanion"
         [zoneOptions]="zoneOptions"
```
sandikbarr, line 71: I don't love passing these all the way down, might just decide to use the store and selectors at the child component level (as long as we don't get too wild with it. it's totally fine, just intentionally deciding where the boundaries are keeps things following the same patterns.)

src/app/campaign/flight/flight-zones-form.component.spec.ts
```diff
@@ -74,21 +76,33 @@ describe('FlightZonesFormComponent', () => {
...
     component.onAddZone();
     expect(component.zones.value).toEqual([
-      { id: '', url: '' },
-      { id: component.zoneOptions[component.flightZones.length].id, url: '' }
+      { id: '', url: '', pingbacks: null },
+      { id: component.zoneOptions[component.flightZones.length].id, url: '', pingbacks: null }
     ]);
   });
 
-  it('removes zones when there are more controls than zones', () => {
```
sandikbarr, line 82: seemed like this test was kind of a lie, which was probably my bad. 😬 fixed

src/app/campaign/flight/flight-form.component.ts
```diff
@@ -56,4 +57,8 @@ export class FlightFormComponent implements OnInit {
   checkError(fieldName: string, type = 'error') {
     return this.flightForm.get(fieldName).getError(type);
   }
+
+  get podcastId() {
+    return this.flight && this.flight.set_inventory_uri && this.flight.set_inventory_uri.split('/').pop();
```
cavis: Hrmm... a bit confusing to refer to this as the "podcast id", since that's actually a different number in our system.  (An Inventory record has an id ... _and_ it has a configured `podcast_id` / `podcast_title`.  Has grove been referring to this as `podcastId` elsewhere already?  Or can we call it a `inventoryId` here?

src/app/campaign/flight/flight-zones-form.component.scss
```diff
@@ -18,3 +18,18 @@
     }
   }
 }
+
+.zone + .zone {
+  border-top: 1px dashed mat-color($prx-gray, 300);
+  padding-top: 20px;
+}
+
+.toggle-pingbacks {
```
cavis, line 25: Noting that this button is a _bit_ taller than the other fields.  And overflow gets a bit weird when width shrinks.

image

But... as this will all move to another "manage creatives" page/UI soon anyway, I don't think it's worth fine-tuning a button that will probably be replaced.

src/app/campaign/flight/flight-zones-form.component.spec.ts
cavis: Good catch!

src/app/campaign/flight/flight-form.error-state-matcher.ts
```diff
@@ -0,0 +1,10 @@
+import { ErrorStateMatcher } from '@angular/material';
+import { FormGroupDirective, FormControl } from '@angular/forms';
+
+export class FlightFormErrorStateMatcher implements ErrorStateMatcher {
```
cavis, line 4: +1 makes sense to have this as a sharable thing.

src/app/campaign/flight/pingbacks/flight-zone-pingbacks-form.component.ts
```diff
@@ -0,0 +1,109 @@
...
+  `,
+  styleUrls: ['./flight-zone-pingbacks-form.component.scss'],
+  providers: [
+    { provide: NG_VALUE_ACCESSOR, useExisting: forwardRef(() => FlightZonePingbacksFormComponent), multi: true },
+    { provide: NG_VALIDATORS, useExisting: FlightZonePingbacksFormComponent, multi: true }
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class FlightZonePingbacksFormComponent implements ControlValueAccessor, OnDestroy {
+  @Input() campaignId: string | number;
```
cavis, line 34: Ah, I see... the campaign is routed through to the pingbacks-preview.  Was wondering where all those inputs/outputs were headed!

src/app/campaign/store/models/flight.models.ts
```diff
@@ -6,6 +6,7 @@ export interface FlightZone {
   id: string;
   label?: string;
   url?: string;
+  pingbacks?: string[];
```
cavis, line 9: One minor issue: if you POST a flight-preview with null pingbacks, Augury currently returns a 500.  (Some piece of the code isn't checking for nil, and doing a `map` on it).  At least that's what I'm seeing everytime I add a new zone to a flight.

For now, may be easiest to either (A) default these to an empty array, the same as flight.targets.  Or (B) don't send the property at all in the POSTed JSON.

Should probably also handle this in Augury, but since pingbacks will all move over to the "Creatives" API ... I'd rather punt until that gets built out.

src/app/campaign/flight/pingbacks/pingback-form.component.ts
```diff
@@ -0,0 +1,196 @@
...
+  previewValue(template: string): string {
+    let preview = '';
+    if (template) {
+      const templateError = this.formGroup.get('template').getError('error');
+      if (templateError) {
+        preview = templateError;
+      } else {
+        preview = template;
+        for (const param of this.findTemplateParams(template)) {
+          if (param === 'campaign' && this.campaignId) {
```
cavis, line 117: Future enhancement: add a bunch of dummy values for the other known template params.

src/app/campaign/flight/pingbacks/pingback-form.component.ts
```diff
@@ -0,0 +1,196 @@
...
+      isDisabled ? this.formGroup.controls[key].disable({ emitEvent: false }) : this.formGroup.controls[key].enable({ emitEvent: false });
+    });
+  }
+
+  validate(_: AbstractControl) {
+    return this.formGroup.valid ? null : { error: 'ERROR: Invalid URL' };
+  }
+
+  findTemplateParams(value: string) {
+    const params = value.split('{').map(v => v.slice(0, v.indexOf('}')));
```
cavis, line 161: Future enhancement: use an npm RFC 6570 library of some sort.  To get the [higher-levels of the RFC](https://tools.ietf.org/html/rfc6570#section-1.2) supported.  (There are some crazy shorthand things you can do).

But... I've never seen ad-ops or a 3rd-party pingback do anything but what you've got here, with the explicit `?param={value}`.

src/app/campaign/flight/pingbacks/flight-zone-pingbacks-form.component.ts
```diff
@@ -0,0 +1,109 @@
...
+            URI syntax help
+          </a>
+        </div>
+      </div>
+    </fieldset>
+  `,
+  styleUrls: ['./flight-zone-pingbacks-form.component.scss'],
+  providers: [
+    { provide: NG_VALUE_ACCESSOR, useExisting: forwardRef(() => FlightZonePingbacksFormComponent), multi: true },
+    { provide: NG_VALIDATORS, useExisting: FlightZonePingbacksFormComponent, multi: true }
```
cavis, line 29: I may have missed this in a past PR... but what exactly do these 2 providers do?  Overriding the accessors/validators on this form-array-ish component?

cavis (changes requested):
One minor issue with the flight-preview (Augury has a bug that doesn't like null `zone.pingbacks`).  But otherwise looks good!

src/app/campaign/flight/flight-form.component.ts
sandikbarr: if this isn't the podcast id, then does it belong in the pingback?

cavis: Oh, I see ... you're pulling this through into the pingback-preview.  Hrmm... so could either pull in the actual `Inventory` record, and lookup the `.podcastId` on that.  Or just don't preview it?  Or throw in a fake number `1234`?  (Any of those would be fine by me).

sandikbarr: ah ok, I'll just take it out for this. We don't use it anywhere else. I just thought I could use it like a feeder url for some reason when I was doing these pingbacks. 🤷

src/app/campaign/flight/pingbacks/flight-zone-pingbacks-form.component.ts
sandikbarr: * NG_VALUE_ACCESSOR registers this ControlValueAccessor with angular as a form control.
* NG_VALIDATORS says it has validation

src/app/campaign/flight/flight-zones-form.component.scss
sandikbarr: I was missing a form-field wrapper to pad that button out. Then I changed the alignment on the grid and took out the flex to center the remove button.

src/app/campaign/store/models/flight.models.ts
sandikbarr: If the pingbacks can be an empty array instead of null, I put in a one line fix for this on the setup of the flight zones form group to use empty array when pingbacks are null/not set.

src/app/campaign/flight/pingbacks/pingback-form.component.ts
sandikbarr: TBH, this preview doesn't really make sense to me because we don't have most of these values to replace in the URL template, and when we move it to being in the creative library we literally only have the creative URL. I think we should discuss this validation using a preview on the creative library when talking to ad ops tomorrow. It doesn't seem all that helpful to me.

cavis: Fair enough.  I'd be fine with just removing the "preview", if the validation message already lists out which/why your url is invalid (checking the params you used).

src/app/campaign/store/models/flight.models.ts
cavis: Fixed!

cavis (approved):
:+100: all looks good to me.  Sounds like we'll try for some feedback on how useful that pingback-preview is later on.  So no use putting more work into it yet!

PR #248, opened 2020-07-27
display geo targets and campaign name on dashboard

### LINKED ISSUE #234: add geo targets to dashboard
on both flights table and campaign cards

### LINKED ISSUE #235: Add campaign name to campaign cards on dashboard
TD;RL: Add the Campaign Name to the the campaign card on the dashboard. Keep the advertiser name with smaller text. Thanks!

from Kali:

[in new system - prod] also just noticed that the campaign names are not listed on the campaign tab of the dashboard and the link that takes you to the campaign is actually the sponsor name. def would like the campaign name on here and would probably make sense for it to be the clickable part that takes you to the campaign. i do like the idea of the sponsor name still being clickable, but taking you to the list of advertiser campaigns

def would like the campaign name on here and would probably make sense for it to be the clickable part that takes you to the campaign. i do like the idea of the sponsor name being here and still being clickable, but taking you to the list of that advertiser's campaigns

sandikbarr:
closes #234 
closes #235


src/app/dashboard/campaign-card/campaign-card.component.ts
```diff
@@ -12,18 +12,19 @@ import { Campaign } from '../dashboard.service';
...
+            {{ campaign.name }}
           </a>
         </h3>
+        <div>{{ campaign.advertiser && campaign.advertiser.label }}</div>
         <div>
           <span class="status {{ campaign.status }}">{{ campaign.status | titlecase }}</span>
           {{ campaign.type | titlecase }}
         </div>
-        <div *ngIf="campaign.flights | campaignFlightTargets as targets">
+        <div *ngIf="campaign.flights | unique: 'targets':'code' | geoTargets as targets">
```
sandikbarr, line 23: I changed these to chain pipes. First one takes the flight targets or zones and creates a unique set based on the key field. The second pipe is shared with the flight table to convert to comma separated labels for display.

cavis: +1 nice!  That `unique` pipe is handy!

cavis (approved):
👍 looks good to me!

PR #249, opened 2020-08-04
campaign csv report

### LINKED ISSUE #242: CSV campaign report
* [ ] Build a campaign CSV download link (with a two dimensional array of the data)
  * The report format has one column per flight listing each flight's parameters and the inventory data for all zones. There are no weekly rollups, just inventory days. It should match the example in the "Vertical" tab here (except in CSV format rather than a google sheet) https://docs.google.com/spreadsheets/d/1yA8cKCANS-V-2Sw4EXPrSWQ-LCx0VocnVfSxMoU6NX0/edit#gid=176452782
* [ ] Add a download CSV report link to the menu in the status bar

sandikbarr:
closes #242

### DIFF
--- a/src/app/campaign/report/campaign-report.component.ts
+++ b/src/app/campaign/report/campaign-report.component.ts
@@ -1,0 +1,78 @@
+import { Component, Input, ChangeDetectionStrategy } from '@angular/core';
+import { DomSanitizer, SafeUrl } from '@angular/platform-browser';
+import { FlightState, Flight } from '../store/models';
+import moment from 'moment';
+
+@Component({
+  selector: 'grove-campaign-report',
+  template: `
+    <a [href]="reportDataCsv" download="{{ reportFilename }}.csv" mat-menu-item>Download CSV</a>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class CampaignReportComponent {
+  @Input() campaignName: string;
+  @Input() flights: FlightState[];
+  reportDataCsv: SafeUrl;
+
+  constructor(private sanitizer: DomSanitizer) {}
+
+  @Input()
+  set reportData(data: any[][]) {
+    const joinedCsvArray =
+      data &&
+      'data:text/csv;charset=utf-8,' +
+        data
+          .map(
+            row =>
+              row &&
+              row
+                .map(v => {
+                  return v && typeof v === 'string' && v.indexOf(',') > -1 ? `"${v}"` : v;
+                })
+                .join(',')
+          )
+          .join('\r\n');
+    this.reportDataCsv = this.sanitizer.bypassSecurityTrustUrl(joinedCsvArray);
+  }
+
+  getFlightDates(flights: Flight[]): { startAt: Date; endAt: Date } {
+    return (
+      flights &&
+      flights.reduce(
+        (acc, flight) => {
+          let startAt = acc.startAt;
+          let endAt = acc.endAt;
+          if (flight.contractStartAt && (!acc.startAt || acc.startAt.valueOf() > flight.contractStartAt.valueOf())) {
+            startAt = new Date(flight.contractStartAt.valueOf());
+          }
+          if (flight.contractEndAtFudged && (!acc.endAt || acc.endAt.valueOf() < flight.contractEndAtFudged.valueOf())) {
+            endAt = new Date(flight.contractEndAtFudged.valueOf());
+          }
+          return { startAt, endAt };
+        },
+        { startAt: null, endAt: null }
+      )
+    );
+  }
+
+  get reportFilename(): string {
+    let fileName = `PRX_${this.campaignName}`;
+    if (this.flights) {
+      const flightDates = this.getFlightDates(this.flights.filter(flight => flight.remoteFlight).map(flight => flight.remoteFlight));
+      if (flightDates.startAt || flightDates.endAt) {
+        fileName += '_';
+      }
+      if (flightDates.startAt) {
+        fileName += moment.utc(flightDates.startAt).format('YYYYMMDD');
+      }
+      if (flightDates.startAt && flightDates.endAt) {
+        fileName += '-';
+      }
+      if (flightDates.endAt) {
+        fileName += moment.utc(flightDates.endAt).format('YYYYMMDD');
+      }
+    }
+    return fileName;
+  }
+}
--- a/src/app/campaign/report/campaign-report.component.ts
+++ b/src/app/campaign/report/campaign-report.component.ts
@@ -43,11 +43,13 @@ export class CampaignReportComponent {
         (acc, flight) => {
           let startAt = acc.startAt;
           let endAt = acc.endAt;
-          if (flight.contractStartAt && (!acc.startAt || acc.startAt.valueOf() > flight.contractStartAt.valueOf())) {
-            startAt = new Date(flight.contractStartAt.valueOf());
+          const flightStartAt = flight.contractStartAt ? flight.contractStartAt : flight.startAt;
+          if (flightStartAt && (!acc.startAt || acc.startAt.valueOf() > flightStartAt.valueOf())) {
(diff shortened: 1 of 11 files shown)


src/app/campaign/store/selectors/campaign-flight.selectors.ts
```diff
@@ -69,3 +76,85 @@ export const selectRoutedCampaignFlightDocs = createSelector(
     return { campaignDoc: campaign.doc, flightDoc: flight.doc };
   }
 );
+
+export const selectCampaignFlightInventoryReportData = createSelector(
+  selectCampaign,
+  selectAllInventory,
+  selectAllAdvertisers,
+  selectRemoteFlightsOrderByContractStartAt,
```
sandikbarr, line 84: using the remoteFlight, can't go back now...

Honestly not sure about this, but I felt like it would be surprising if a report had unsaved data. However, we overwrite the inventory flight days with the preview data instead of storing the preivew separately. The actuals shouldn't change in the preview, but it could still be weird/off.

src/app/campaign/report/campaign-report.component.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  getFlightDates(flights: Flight[]): { startAt: Date; endAt: Date } {
+    return (
+      flights &&
+      flights.reduce(
+        (acc, flight) => {
+          let startAt = acc.startAt;
+          let endAt = acc.endAt;
+          if (flight.contractStartAt && (!acc.startAt || acc.startAt.valueOf() > flight.contractStartAt.valueOf())) {
```
cavis: Maybe we should also check `flight.startAt` if `flight.contractStartAt` is null?  (Just ran a test export locally, and most of my flights have no contract dates, since they're optional fields).

src/app/campaign/store/selectors/campaign-flight.selectors.ts
cavis: Agreed - it would be surprising for anything unsaved to show up.  Maybe that's an argument to store both local/preview and remote/persisted "flight-days"?  Such an edge case though ... I don't think it's worth addressing yet.

src/app/campaign/store/selectors/campaign-flight.selectors.ts
```diff
@@ -69,3 +76,85 @@ export const selectRoutedCampaignFlightDocs = createSelector(
...
+          const advertiser = advertisers.find(a => a.set_advertiser_uri === campaign.remoteCampaign.set_advertiser_uri);
+          return advertiser ? advertiser.name : '';
+        })
+      )
+    );
+    // Zones
+    reportData.push(['Zones'].concat(flights.map(flight => flight.zones.map(zone => zone.label).join('|'))));
+    // Contract Start and End Dates
+    reportData.push(
+      ['Start Date'].concat(flights.map(flight => (flight.contractStartAt ? flight.contractStartAt.format('MM-DD-YYYY') : '')))
```
cavis: Yeah, and same here - I think these should fallback to the non-contract start/end dates.

src/app/campaign/store/selectors/campaign-flight.selectors.ts
```diff
@@ -69,3 +76,85 @@ export const selectRoutedCampaignFlightDocs = createSelector(
...
+        flights.map(flight =>
+          flight.targets
+            .map(t => geoTransform.transform(t))
+            .filter(t => t)
+            .join('|')
+        )
+      )
+    );
+    // Contract Goal
+    reportData.push((['Contract Goal'] as (string | number)[]).concat(flights.map(flight => flight.contractGoal)));
```
cavis: And fallback to the actual totalGoal if null.

src/app/campaign/store/selectors/campaign-flight.selectors.ts
```diff
@@ -69,3 +76,85 @@ export const selectRoutedCampaignFlightDocs = createSelector(
...
+      {}
+    );
+    // push each inventory date row
+    // if no data for flight on date, show '-'
+    Object.keys(inventoryDays)
+      .sort((a, b) => new Date(a).valueOf() - new Date(b).valueOf())
+      .forEach(date =>
+        reportData.push(
+          ([utc(date).format('MM-DD-YYYY')] as (string | number)[]).concat(
+            flights.map(flight => (inventoryDays[date][flight.id] ? inventoryDays[date][flight.id].numbers.actuals : '-'))
```
cavis, line 168: I don't love that in-flight-range days show up as blanks in the CSV instead of 0s.  But I can't remember what ad-ops actually said about that ... so I'm fine with just getting this out there for some feedbacks!

cavis (changes requested):
I do think we should fallback to non-contract start/end/goal fields when the contract ones are null.  Just to reinforce that they're optional.  But other than that this looks good!

src/app/campaign/report/campaign-report.component.ts
sandikbarr: Yeah, they said they hadn't filled most of the contract dates out. It confused me that this is what they wanted in the report.

Ok, so the fields are independent... So what if there is a contract start but not a contract end date, but actual start and end date are filled out as those are required. What should it display then? Should the contract start date be mixed with the actual end date? Honestly it might make more sense to list both dates separately as Kali was suggesting yesterday.

src/app/campaign/store/selectors/campaign-flight.selectors.ts
sandikbarr: Same question here about falling back to both actual dates if the contract start or end date is missing, or fallback independently.

src/app/campaign/store/selectors/campaign-flight.selectors.ts
sandikbarr: In the API, dates with no actuals are null, that's why you're seeing them as blanks instead of '-' where there is no expected data. I think they are expecting there to be actuals during the flight dates, so I'm not actually sure what they want when there isn't or if the flight hasn't finished. Yeah, feedback would be good.

src/app/campaign/report/campaign-report.component.ts
cavis: Good questions!

My perhaps bad assumption is: if you leave contract-start or contract-end blank, they "default" to the start or end dates.  At least as far as reporting is concerned!  So here i'd expect these to be sort of coalesced fields... `startAt: flight.contractStartAt || flight.startAt` (and the same with endAt).

But I would also be fine with just displaying them both.  More data, more clarity!

src/app/campaign/store/selectors/campaign-flight.selectors.ts
cavis: I was thinking fallback independently.

Per the last ad-ops meeting ... I do think we'll start defaulting `f.contractStartAt = f.startAt` and `f.contractEndAt = f.endAt` on the frontend at some point.  Which sort of makes them non-optional fields.  But doesn't hurt to belt-and-suspenders to do the fallback here.

src/app/campaign/store/selectors/campaign-flight.selectors.ts
cavis: Ah, okay.  I do remember them wanting dashes.  To me I guess `0` is more indicative of "this flight _could have_ served this day, but we got 0 actuals".  And `-` sort of means "not available".  But we can get feedback on that!

cavis (approved):
👍 yep, that looks good to me now!

src/app/campaign/store/selectors/campaign-flight.selectors.ts
sandikbarr: If it _could_ have served, like if the date is in the past and there are still no actuals, the API probably really should have zero instead of null. null makes sense to me for dates in the future, or if they really do want '-' filled in there because they are exporting the report before the end of the flight. There is no differentiation of when the flight should be running or not if we just fill everything empty with '-'. But hey, let's not decide requirements in the PR, not the right place.

PR #251, opened 2020-08-06
Rename allocation status

cavis:
Closes PRX/augury.prx.org#301.

Renames flight `status` and `statusMessage` to `allocationStatus` and `allocationStatusMessage`.  In prep for moving the current campaign-status field to flight-status.

Must deploy concurrently with PRX/augury/.prx.org#321.


sandikbarr (approved):
lgtm!

PR #259, opened 2020-08-24
Flight status

### LINKED ISSUE #246: Implement Flight statuses
See #206.  After PRX/augury.prx.org#303.

- [x] Remove campaign status field from the campaign form
- [x] Add a flight status field to the flight form before the zones (after the series selection). Simple selectbox. Same states as the campaign had
- [x] Replace the campaign status with the flight state in the flight dashboard.
- [x] Remove campaign status from the campaign cards.
- [x] TODO: what about the dashboard filters/facets?
  - PRX/augury.prx.org#303 will add a "flight status" filter/facet.  But the campaign status filter/facet will disappear.
  - Maybe the status filter isn't valid on the campaign dashboard?

cavis:
Closes #246.  Requires and should be deployed concurrently with PRX/augury.prx.org#331.

Move the campaign status dropdown over to the flight form.  Fixup dashboard.

image


sandikbarr (approved):
Looks good. There's some status CSS leftover, but it might come up again. Anything else I would've changed or cleaned up was originally code I likely wrote and looks like you copied over. 😬  I'm just going to believe this works until we can get these two PRs with model changed merged together.

Issue #236, opened 2020-07-21
Creative Management


brandonhundt:
Ability to manage creatives separate from the Flight.

Will include the following:
- Creative Dashboard
- Creative Add/Edit Form
- Ability to add/edit creatives from the Flight form and assign weights (weighting will be stored on the flight)

Creatives are now cards
* When a user clicks "Add new creative" I'd like to use the pattern set up with targets. User can then quickly select the type of creative they want to use (add new, existing, silent file)
* If silent is selected, a silent card is added straight away, bypassing the flyout (see the third example)
New File Flyout (first image)
* Added Advertiser field to new file flyout
* Added link to look for an existing file
Existing File flyout (2nd image):
* Filters (with advertiser defaulted to the campaign advertiser)
* Shows 5 latest creatives in a condensed card layout.

![DT - Creatives - Full Add new](https://user-images.githubusercontent.com/2250413/89921021-c8f72580-dbc2-11ea-9b56-885fb3542670.jpg)
![DT - Creatives - Full -existing](https://user-images.githubusercontent.com/2250413/89921030-cd234300-dbc2-11ea-9eb9-ec619e6bd1b0.jpg)

PR #265, opened 2020-09-11
multiple flight zone creatives and creative library

### LINKED ISSUE #252: Creative Management Flight Slideout Form
Add/Edit Creatives on a Flight Zone with a slideout form on the side

### LINKED ISSUE #253: Flight Zone Creative Card
Creatives on the Flight Zones shown as a card

sandikbarr:
closes #252 and closes #253


src/app/campaign/creative/creative-list.component.ts
```diff
@@ -0,0 +1,137 @@
...
+        </mat-list-option>
+      </mat-selection-list>
+      <mat-paginator
+        [length]="creativeTotal$ | async"
+        [pageIndex]="params.page - 1"
+        [pageSize]="params.per"
+        [pageSizeOptions]="[5, 10, 20]"
+        (page)="onPage({ page: $event.pageIndex + 1, per: $event.pageSize })"
+      ></mat-paginator>
+      <div class="form-submit">
```
sandikbarr, line 57: Update: Realized I forgot to add the Save/Cancel buttons as shown in the design. https://github.com/PRX/grove.prx.org/issues/236 When I've shown this in demos, I had it so you just clicked to select a single creative and it would be added to the zone and the drawer closed without any additional buttons. The MatSelectionList supports single selection, but it still looks like a checkbox rather than a radio button as we have in the design. Having the save button down below also made me think choosing multiple creatives and clicking a button would be more appropriate, and of course that had to work across this paged list.

src/app/core/creative/creative.service.ts
```diff
@@ -0,0 +1,52 @@
...
+    text?: string;
+    sorts?: string;
+    direction?: string;
+  }): Observable<{ creativeDoc: HalDoc; advertiserDoc: HalDoc }[]> {
+    return this.augury
+      .followItems('prx:creatives', {
+        ...params,
+        zoom: 'prx:advertiser',
+        sorts: `${params.sorts}:${params.direction}`,
+        ...(params.text && { filters: `text:${params.text}` })
```
cavis: Hmm... I think filters are formatted with `name=value` not `name:value`.  This doesn't seem to return any different result.

src/app/campaign/store/reducers/flight.reducer.ts
```diff
@@ -137,11 +139,25 @@ const _reducer = createReducer(
...
+    const { flightId, zoneId, creativeIds } = action;
+    const localFlight = state.entities[flightId] && state.entities[flightId].localFlight;
+    const creativeFlightZones = creativeIds.map(creativeId => ({ creativeId }));
+    const zones =
+      localFlight &&
+      localFlight.zones.map(zone =>
+        zone.id === zoneId
+          ? {
+              ...zone,
+              creativeFlightZones: zone.creativeFlightZones ? [...zone.creativeFlightZones, ...creativeFlightZones] : creativeFlightZones
```
cavis, line 153: Okay, so this is where select creative-ids are appended to the zone.  (With no deduping, since you're allowed to reuse the creative in the zone more than once).

src/app/campaign/store/reducers/creative.reducer.ts
```diff
@@ -0,0 +1,87 @@
...
+      })),
+      { ...state, params: action.params, total: action.total }
+    )
+  ),
+  on(
+    creativeActions.CreativeLoadFailure,
+    creativeActions.CreativeCreateFailure,
+    creativeActions.CreativeUpdateFailure,
+    creativeActions.CreativeLoadListFailure,
+    (state, action) => ({ ...state, error: action.error })
```
cavis, line 75: Should I be seeing these bubble up to the UI somewhere?  When I try to create/update a bad creative (set the url to something non-existent) ... I don't get an error message/toast anywhere.

src/app/campaign/creative/creative-list.component.ts
cavis: +1.  And paging does complicate things.  /me tries to wrap head around those codes...

src/app/campaign/creative/creative-list.component.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  onPage({ page, per }: { page: number; per: number }) {
+    this.store.dispatch(creativeActions.CreativeLoadList({ params: { page, per } }));
+  }
+
+  onSave() {
+    this.campaignId$.pipe(withLatestFrom(this.flightId$, this.zoneId$), first()).subscribe(([campaignId, flightId, zoneId]) => {
+      this.router.navigate(['/campaign', campaignId, 'flight', flightId]);
+      const creativeIds = Object.keys(this.selectedCreativeIds)
+        .map(page => this.selectedCreativeIds[page])
```
cavis, line 119: Let's see if I'm following this at all ...

So you're keeping track of the selected creative ids, as a hash of `page => [ids]` ... in `this.selectedCreativeIds`.

And you have to keep those page numbers in the state because ... you have to select the checkboxes across multiple pages?

src/app/campaign/flight/flight-form.component.spec.ts
```diff
@@ -120,16 +125,20 @@ describe('FlightFormComponent', () => {
...
-        FlightZonesFormComponent,
-        FlightZonePingbacksFormComponent,
-        PingbackFormComponent
+        MatMenuModule,
+        StoreModule.forRoot(
+          { router: routerReducer },
+          {
+            runtimeChecks: {
+              strictStateImmutability: true,
+              strictActionImmutability: true
```
cavis, line 134: These are the defaults, yes?  I'm guessing this is silencing some deprecation warning or something then.

src/app/campaign/flight/flight.container.ts
```diff
@@ -85,11 +88,38 @@ export class FlightContainerComponent implements OnInit, OnDestroy {
     this.targetOptionsMap$ = this.store.pipe(select(selectCurrentInventoryTargetsTypeMap));
     this.flightActualsDateBoundaries$ = this.store.pipe(select(selectFlightActualsDateBoundaries));
     this.campaignId$ = this.store.pipe(select(selectCampaignId));
+
+    // subscribe to creativeId route param/creative list route and open drawer
```
cavis, line 92: +10 good solution!  (Triggering drawer open/closed based on the route).

src/app/campaign/store/actions/creative-action.creator.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const CreativeFormUpdate = createAction(
+  ActionTypes.CREATIVE_FORM_UPDATE,
+  props<{ creative: Creative; changed: boolean; valid: boolean }>()
+);
+
+// could get the flightId into this action, but don't have the zone --> will also have to add zone to route
+export const CreativeCreate = createAction(
+  ActionTypes.CREATIVE_CREATE,
+  props<{ campaignId?: string | number; flightId?: number; zoneId?: string; creative: Creative }>()
```
cavis, line 20: Wondering if this will slightly conflict with standalone creative-creation (outside of a flight) in PRX/internal#584.  I suppose the flight/zone are optional ... so maybe not.

src/app/campaign/store/effects/creative.effects.ts
```diff
@@ -0,0 +1,112 @@
...
+            mergeMap(creativeDoc => {
+              const { campaignId, flightId, zoneId, creative } = action;
+              if (campaignId && flightId) {
+                // route back to flight (if there is one in the original action)
+                this.router.navigate(['/campaign', campaignId, 'flight', flightId]);
+              }
+              // save creative success and save creative to flight
+              return [
+                creativeActions.CreativeCreateSuccess({ campaignId, flightId, zoneId, creativeDoc }),
+                campaignActions.CampaignFlightZoneAddCreatives({ flightId, zoneId, creativeIds: [creativeDoc.id] })
```
cavis, line 46: Ah, I see ... adding the new creatives to the flight is its own action.  +yes.  :+i-get-things:

src/app/campaign/store/effects/creative.effects.ts
```diff
@@ -0,0 +1,112 @@
...
+              if (campaignId && flightId) {
+                // route back to flight (if there is one in the original action)
+                this.router.navigate(['/campaign', campaignId, 'flight', flightId]);
+              }
+              return creativeActions.CreativeUpdateSuccess({ campaignId, flightId, zoneId, creativeDoc });
+            }),
+            tap(_ => {
+              this.toastr.success('Creative saved');
+            }),
+            catchError(error => of(creativeActions.CreativeUpdateFailure({ error })))
```
cavis, line 79: No toast on these?

src/app/campaign/store/models/flight.models.ts
```diff
@@ -1,14 +1,12 @@
...
+import { Creative } from './creative.models';
 
 export interface FlightZone {
   id: string;
   label?: string;
-  url?: string;
-  pingbacks?: string[];
-  fileSize?: number;
-  mimeType?: string;
+  creativeFlightZones?: { creativeId?: number | string; weight?: number; disabled?: boolean }[];
```
cavis, line 9: Wait ... so you're not even using the `creativeFlightZones[].id`?  Why does the API even include it then?  May want to circle back on that after this merges ...

cavis (changes requested):
Phew!  Couple of minor things (text filter not working; save-creative errors not surfacing).  But overall this is looking great!  Excited to merge and iterate!

There is some missing functionality around disabling creatives, the "draft" workflow, and even deleting/removing creatives entirely prior to the flight being approved/live.  Much of that is already captured in tickets/epics.

src/app/campaign/creative/creative-list.component.ts
sandikbarr: The MatSelectList is not the most user friendly api. Angular Material's docs in general are about as good as the rest of their docs, very limited information. This MatSelectList doesn't know about the paging, so I've got to keep track of those and reset them between pages. So on selectionChange, it sets the selected list items from the current page. Then when the items are rendered, selected is set according to what's in the `selectedCreativeIds: {[page: number]: number[]}`

src/app/campaign/creative/creative-list.component.ts
sandikbarr: yes, that's it exactly.

src/app/campaign/flight/flight.container.ts
sandikbarr: I was kinda worried folks would think this is wonky, but you can route the app directly to the creative library or form, so it works.

src/app/campaign/store/actions/creative-action.creator.ts
sandikbarr: Yes, that is why the campaignId and flightId are optional. Basically, if they are not present, it didn't come from the campaign/flight route, so it won't route back to it on success. [Mike Ryan might disagree with me about ngrx action hygiene and say we shouldn't expect to reuse the action in another part of the app, but we can decide at that time if we really think the app is so big that we'd get confused about where a CreativeCreate action came from.]

src/app/campaign/store/effects/creative.effects.ts
sandikbarr: woo! yes! I love it when things click and start making sense!

src/app/campaign/store/effects/creative.effects.ts
sandikbarr: I don't think errors belong in a toast that will disappear on it's own. The fact that it pops up and goes away to me means it is intended to be supplemental information. We really do need to discuss how we should be handling alerting and errors in the application. I've done what I can with it (when I can actually get permission to do a thing or just slide it on in with something else, and if you don't believe me we can talk about how many 1:1s it took for me to get to do the dismissible snackbar error on campaign/flight save.) IMO this has been a product TODO and needs to be addressed by design. I never intended to keep the toast when I started adding it to this app. It was always a question of what are we intending to do with alerts and errors. Grove #271

src/app/campaign/store/models/flight.models.ts
sandikbarr: Nope, that unique id is not used. The creatives on the flight are acted on as a whole.

src/app/campaign/store/reducers/creative.reducer.ts
sandikbarr: Hey now, don't go adding new requirements in my PR. #271

src/app/campaign/store/reducers/flight.reducer.ts
sandikbarr: I went through an iteration with this where the creatives were including in the flight zones and I was just storing them here as they came in so I could use the creative filename for the card. I had to do some "creative" effects handling to get those creatives loaded with the flights, but it's much better. I also considering moving flight zones to a separate reducer or at least changing the shape of them here so they are keyed by id instead of an array, but I decided to stop where I had things and get this done.

sandikbarr (commented):
created #271 for creatives errors that we want shown to the user, and hopefully I got all the rest of your questions and comments @cavis

src/app/campaign/store/reducers/creative.reducer.ts
cavis: +1 makes good sense!

cavis (approved):
👍 looks good to me!

I also added a comment in PRX/augury.prx.org#342 to maybe just remove the creative flight zone id.

PR #267, opened 2020-09-21
Fixes duplication issues

### LINKED ISSUE #254: Campaign Duplication Error
Attempting to duplicate the campaign from the menu in the status bar on the form page now throws an error why it tries to serialize the campaign and flights into the route data and fails with error. Users can work around this error by using the other campaign duplication method via the campaign dashboard (not sure how much they use that campaign cards view.)

Screen Shot 2020-08-13 at 1 45 14 PM

### LINKED ISSUE #255: Clear Flight Dates on Duplication
The startAt and endAt fields should be empty when duplicating a campaign or flight. We need to update the support for this for the fudged end at fields and probably add the contract start and end dates as well.

sandikbarr:
Two for one!
* closes #254 resolves the campaign duplication error by removing Moment flight dates that don't serialize on router state
* closes #255 clears _all_ flight dates on duplication

### DIFF
--- a/src/app/campaign/store/effects/campaign.effects.ts
+++ b/src/app/campaign/store/effects/campaign.effects.ts
@@ -259,8 +259,15 @@ export class CampaignEffects {
             campaign,
             flights: flights.map(flight => ({
               ...flight,
+              // TWO THINGS:
+              // * Clear all flight dates on duplication
+              // * Moments cannot be serialized into state on the router, so the dates error unless translated to a string type or whatever
               startAt: undefined,
-              endAt: undefined
+              endAt: undefined,
+              endAtFudged: undefined,
+              contractStartAt: undefined,
+              contractEndAt: undefined,
+              contractEndAtFudged: undefined
             }))
           };
           this.router.navigate(['/campaign', 'new'], { state });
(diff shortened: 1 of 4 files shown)


src/app/campaign/store/effects/campaign.effects.ts
```diff
@@ -259,8 +259,15 @@ export class CampaignEffects {
             campaign,
             flights: flights.map(flight => ({
               ...flight,
```
sandikbarr, line 261: Two for one! How convenient! Clearing the flight dates is a business rule that got left behind when we added contract dates and also did the fudged date fields, and it was those Moments causing the error. If we want the dates, we can't pass them through this router state as Moments, they would need to be strings or something.

src/app/campaign/store/reducers/flight.reducer.ts
```diff
@@ -68,7 +68,7 @@ const _reducer = createReducer(
   }),
   on(campaignActions.CampaignDupFlight, (state, action) => {
     const { flight, flightId } = action;
```
sandikbarr, line 70: bonus fix: duplicated flight names were getting set to "Copy of XYZ (Copy)"

cavis (approved):
👍  looks good to me!  Feel free to merge.

PR #272, opened 2020-09-24
save accumulator reference in reduce to improve performance

sandikbarr:
Improves performance of CSV report generation. If a flight date gets set far out into the future, that flight's inventory table still takes a while to generate, but this fix at least makes the other flight pages usable.

### DIFF
--- a/src/app/campaign/store/selectors/campaign-flight.selectors.ts
+++ b/src/app/campaign/store/selectors/campaign-flight.selectors.ts
@@ -12,7 +12,7 @@ import {
 } from './flight.selectors';
 import { selectFlightDaysEntities } from './flight-days.selectors';
 import { selectAllInventory } from './inventory.selectors';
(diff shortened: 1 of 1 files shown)


src/app/campaign/store/selectors/campaign-flight.selectors.ts
```diff
@@ -147,24 +147,22 @@ export const selectCampaignFlightInventoryReportData = createSelector(
...
-            }
+          const date = day.date.valueOf();
+          acc[date] = {
+            ...acc[date],
+            [flight.id]: day
           };
+          return acc;
         }, days),
-      {}
+      {} as { [date: number]: { [flightId: number]: FlightDay } }
```
sandikbarr, line 157: This code can be hard to read and understand. Sorry I didn't add this typing in here to begin with. It does seem to help to know what the heck it's trying to build here.

cavis (approved):
👍 looks good to me!  Javascript!  Bwah!

PR #279, opened 2020-09-29
don't disable creative weight control on creative disable

sandikbarr:
to be deployed after augury disable creatives PR https://github.com/PRX/augury.prx.org/pull/351

### DIFF
--- a/src/app/campaign/creative/creative-card.component.scss
+++ b/src/app/campaign/creative/creative-card.component.scss
@@ -17,6 +17,11 @@ h3 {
   justify-content: space-between;
 }
 
+.creative.disabled {
+  color: prx-theme-foreground(disabled-button);
+  text-decoration: line-through;
+}
+
 .creative {
   display: block;
   max-width: 100%;
--- a/src/app/campaign/creative/creative-card.component.ts
+++ b/src/app/campaign/creative/creative-card.component.ts
@@ -1,6 +1,5 @@
 import { Component, Input, OnInit, OnDestroy, ChangeDetectionStrategy } from '@angular/core';
 import { FormGroup, ControlContainer } from '@angular/forms';
-import { Subscription } from 'rxjs';
 import { Creative } from '../store/models';
 
 @Component({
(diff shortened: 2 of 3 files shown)


src/app/campaign/creative/creative-card.component.scss
```diff
@@ -17,6 +17,11 @@ h3 {
   justify-content: space-between;
 }
 
+.creative.disabled {
+  color: prx-theme-foreground(disabled-button);
```
sandikbarr, line 21: I dunno 🤷 , I thought disabled cards should have more of a different look than just the toggle. Initially I disabled the weight control, but that turned out to be wrong because the weight is still required and disabling a form control means it isn't included in the form's valueChanges updates, so disabled creatives couldn't have a weight.

This color does not have enough contrast on the light background, but it is using the material theme.

cavis: :+12: i thought the strikethrough made it pretty obvious what was happening.  Silent Files don't get the same treatment ... but those are an edge-case in real life anyways.

src/app/campaign/creative/creative-card.component.ts
```diff
@@ -21,28 +22,18 @@ import { Creative } from '../store/models';
...
-export class CreativeCardComponent implements OnInit, OnDestroy {
+export class CreativeCardComponent implements OnInit {
   @Input() creative: Creative;
   @Input() creativeLink: string;
   creativeForm: FormGroup;
-  disableChangedSubscription: Subscription;
 
   ngOnInit() {
     this.creativeForm = this.formContainer.control as FormGroup;
-    this.disableChangedSubscription = this.creativeForm.get('enabled').valueChanges.subscribe(enabled => this.setDisabledState(!enabled));
```
cavis, line 32: Ah, okay, so this isn't needed now that the form doesn't "react" to the toggle changing.  (You don't actively disable the other field).

cavis (approved):
👍 Looks good to me!

PR #282, opened 2020-10-08
angular 9 upgrade

### LINKED ISSUE #263: Angular 9 upgrade
upgrade Grove application to Angular 9

sandikbarr:
closes #263


angular.json
```diff
@@ -17,6 +17,7 @@
         "build": {
           "builder": "@angular-devkit/build-angular:browser",
           "options": {
```
sandikbarr, line 19: aot build and using Ivy by default

src/app/app.module.ts
```diff
@@ -51,8 +51,8 @@ const PREFERRED_DATE_FORMATS = {
     BrowserAnimationsModule,
     DashboardModule,
     MatMomentDateModule,
-    StoreModule.forRoot(reducers, { metaReducers }),
-    StoreModule.forRoot({ router: routerReducer }),
```
sandikbarr, line 55: _sigh_ seems there's some kind of mutation of the state somewhere and you can't load up the campaign feature (campaign/flight form) without running into a "Cannot freeze" error. These runtime checks get turned on by default in ngrx 9.

src/app/campaign/campaign.component.spec.ts
```diff
@@ -2,7 +2,11 @@ import { TestBed, async, ComponentFixture } from '@angular/core/testing';
 import { DebugElement } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 import { RouterTestingModule } from '@angular/router/testing';
-import { MatIconModule, MatListModule, MatMenuModule, MatProgressSpinnerModule, MatSidenavModule } from '@angular/material';
```
sandikbarr, line 5: There's an automated migration that updates the material imports for better tree shaking.

src/app/campaign/campaign.routing.ts
```diff
@@ -73,4 +73,4 @@ export const campaignComponents: any[] = [
   CreativeCardComponent
 ];
 
-export const campaignRouting: ModuleWithProviders = RouterModule.forChild(campaignRoutes);
```
sandikbarr, line 76: ModuleWithProviders needs a type now.

src/app/campaign/store/reducers/campaign.reducer.spec.ts
```diff
@@ -10,6 +10,7 @@ import {
   flightDocFixture,
   flightDaysDocFixture
 } from '../models/campaign-state.factory';
+import { CampaignState } from '../models';
```
sandikbarr, line 13: some stricter type checking, but we didn't get caught too bad

src/app/campaign/store/effects/advertiser.effects.ts
```diff
@@ -10,18 +10,24 @@ export class AdvertiserEffects {
   loadAdvertisers$ = createEffect(() =>
     this.actions$.pipe(
       ofType(advertiserActions.AdvertisersLoad),
-      switchMap(() => this.advertiserService.loadAdvertisers()),
-      map(docs => advertiserActions.AdvertisersLoadSuccess({ docs })),
-      catchError(error => of(advertiserActions.AdvertisersLoadFailure({ error })))
+      switchMap(() =>
```
sandikbarr, line 13: Probably my fault for doing it this way with flattened streams initially all over in the Effects. Error handling in Effects should be nested within a higher order operator or the effect stream will complete.  https://ngrx.io/guide/effects#handling-errors You can see in the marble diagrams where I took out the `(a|)` (action+completion grouping) in the tests that the effect will not complete on error, which is what we want. Ngrx 9 changed effects to only retry a default of 10 times when it resubscribes on error, so it seemed best to get these right. https://ngrx.io/guide/migration/v9#limit-retries-to-10-by-default

src/app/campaign/store/reducers/account.reducer.ts
```diff
@@ -14,7 +14,7 @@ export const initialState: State = adapter.getInitialState({});
 // tslint:disable-next-line: variable-name
 const _reducer = createReducer(
   initialState,
-  on(accountActions.AccountsLoadSuccess, (state, action) => ({ ...adapter.addAll(action.docs.map(docToAccount), state), error: null })),
```
sandikbarr, line 17: addAll is deprecated in favor of setAll (better naming)

src/app/app.module.ts
rpeterman-gp: Is this something we would want to fix in the future?

rpeterman-gp (approved):
Approved with question regarding added runtime checks.

src/app/app.module.ts
sandikbarr: Probably yes because it could potentially lead to weird bugs. It could be hard to track down, but there are freeze utilities you might be able to apply to individual reducers to find where there is mutation happening. My guess is that it's in the flight reducer since that's a deep object, but I didn't spend enough time tracking it down. I did try the ngrx-store-freeze application wide, and I got the same error.

PR #283, opened 2020-10-13
use validations on startAt and endAt instead of min/max

### LINKED ISSUE #256: Datepicker min/max
The min/max setting that disabled dates in the datepickers where start > end or end < start are clunky with start/end range selection. Keep the validation and show an error if start > end, but allow users to freely select the dates so they can more easily choose start and end independently for the date range.

Angular 10 will add support for an Angular Material Date Range component and should generally improve date range picking UX.

sandikbarr:
closes #256


src/app/campaign/flight/flight-form.component.ts
```diff
@@ -26,7 +26,6 @@ export class FlightFormComponent implements OnInit {
   @Output() flightDeleteToggle = new EventEmitter(true);
 
   flightForm: FormGroup;
-  matcher = new FlightFormErrorStateMatcher();
```
sandikbarr, line 29: this doesn't apply anymore since we changed datepickers to update on blur, removed

rpeterman-gp (approved):
Looks great!

Are there plans to disable dates in the calendar widgets that would result in invalid input? Like disable the dates after the selected end date in the start date widget.

sandikbarr:
> Are there plans to disable dates in the calendar widgets that would result in invalid input? Like disable the dates after the selected end date in the start date widget.

This is actually what I removed with the min/max on the date inputs. There was feedback from Madeline and Hunter that it was annoying to set flight dates.

There is a date range input in angular material 10 though that I think will improve these start/end dates altogether.

PR #284, opened 2020-10-16
saves flight haldoc into state on create flight

### LINKED ISSUE #280: Create-then-Update flight doesn't fire
Something breaking down if you create a flight then update it without any other navigation/refreshing.

Steps:

- Navigate to one of your campaigns
- Add a new flight.  Set status=hold, any series, any zone, blank creative, weight=1.
- Click save - see the PUT-campaign and POST-flight fire.
- Change status=sold.
- Click save - see PUT-campaign, but nothing else.

Expected:

- Flight should PUT on that 2nd save.

sandikbarr:
closes #280

### DIFF
--- a/src/app/campaign/store/reducers/flight.reducer.ts
+++ b/src/app/campaign/store/reducers/flight.reducer.ts
@@ -125,7 +125,7 @@ const _reducer = createReducer(
       newState = adapter.addMany(
         newFlightIds.map(newId => {
           const flight = docToFlight(createdFlightDocs[newId]);
-          return { localFlight: flight, remoteFlight: flight, changed: false, valid: true };
+          return { doc: createdFlightDocs[newId], localFlight: flight, remoteFlight: flight, changed: false, valid: true };
         }),
         newState
       );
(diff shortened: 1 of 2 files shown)


src/app/campaign/store/reducers/flight.reducer.ts
```diff
@@ -125,7 +125,7 @@ const _reducer = createReducer(
       newState = adapter.addMany(
         newFlightIds.map(newId => {
           const flight = docToFlight(createdFlightDocs[newId]);
-          return { localFlight: flight, remoteFlight: flight, changed: false, valid: true };
+          return { doc: createdFlightDocs[newId], localFlight: flight, remoteFlight: flight, changed: false, valid: true };
```
sandikbarr, line 128: The campaign.service gets the haldoc from the store in order to call update against it. https://github.com/PRX/grove.prx.org/blob/master/src/app/core/campaign/campaign.service.ts#L82
But the CampaignSaveSuccess was not saving the haldoc with created flights, whoops!

Fixed _forever_

cavis (approved):
:+huzzah!:

PR #286, opened 2020-10-16
fixes spec warnings: Navigation triggered outside Angular zone

sandikbarr:
just a bonus fix on some navigation warnings in these specs


cavis (approved):
👍

Issue #288, opened 2020-10-20
Flight creatives styling


cavis:
[Style is important.](https://www.sketch.com/s/add11476-b42d-44cc-a705-8d2bde1f3f69/a/848yvY)

- [x] Update styles
- [x] Set a default weight, and make it a percentage.
- [ ] Allow removing creatives when they have no actuals.  Otherwise, you can only disable.
- [x] Make sure approved flights have at least 1 enabled creative.
- [ ] Show creative actuals, if those have made it into the API already.
- [x] Allow non-approved flights to exist without creatives (may require changes to Augury to relax that validation).

![image](https://user-images.githubusercontent.com/1410587/96608934-2be9f480-12b7-11eb-9544-bda2569714c3.png)

PR #289, opened 2020-10-26
Creative styles/validations

cavis:
Fixes most of #288.  (CreativeFlightZone actuals aren't in the API yet, so still 2 TODOs there).

Deviated _a little bit_ from the designs, to minimize vertical space usage.  But basically the same:

image

Also changed the "zone name" dropdown to bring this into parity with the targets picker.  So you have to select a zone-name from the menu at the bottom.  (I think this is slightly more intuitive than being able to change the name of an existing zone).

image


svevang (approved):
Looks good!

PR #294, opened 2020-10-28
Competing flights

### LINKED ISSUE #274: Competing Flights UI
Fetch and display competing flights under the Inventory table.

image

- [x] Fetch overlapping flights using the list-flights API:
  - Use the current flight form start/end dates, inventory (podcast id), and zone name.
  - `GET /api/v1/flights?zoom=0&per=999&sorts=allocationPriority:desc&filters={podcast,zone,before,after}`
- [ ] Add a collapsible (open by default) "competing flights" section from the design.
- [ ] Make the inventory table itself also collapsible (open by default).
- ~[ ] Have the flight links open in new tabs.~ _**Update**_: respect browser settings. Let's rely on #77 to prevent navigating away from an unsaved form.
- [x] Make sure the current flight isn't displayed in the list.

cavis:
Closes #274.  Displays flights that _might_ compete with this one, based on inventory/zones/start/end.

Fallout: #296 does not do the expand/collapse panels.  I'll open a separate ticket to address that.  Requires some re-theming of the Material expansion-panel that gets tricky.


src/app/campaign/inventory/inventory.component.scss
```diff
@@ -1,10 +1,5 @@
 @import '~src/sass/variables';
 
-$row-line-height: 20px;
```
cavis, line 3: These variables all moved to the `inventory-table.component.scss`.  So weren't actually doing anything here.

src/app/campaign/inventory/inventory.component.scss
```diff
@@ -29,3 +24,49 @@ section {
...
+
+.empty {
+  height: 80px;
+  display: grid;
+  align-items: center;
+  justify-content: center;
+  color: mat-color($prx-gray, 400);
+}
+
+.row {
```
cavis, line 47: This is a bit duplicative of the inventory-table css ... but it's non trivial to share that between components, or at least i don't know a great way.  Ideally, we'd convert these all to Material tables, and restyle from there.

src/app/campaign/store/effects/flight-overlap.effects.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor(
+    private actions$: Actions<flightOverlapActions.FlightOverlapActions>,
+    private store: Store<any>,
+    private overlap: FlightOverlapService
+  ) {}
+
+  // only reload overlapping flights when the filter params have changed
+  private shouldLoadOverlap([flight, currentFilters]: [Flight, string]) {
+    const newFilters = flight && overlapFilters(flight);
+    return newFilters && newFilters !== currentFilters;
```
cavis, line 55: This is the magic ... that `overlapFilters` function calculates the `?filters=` string from a `Flight`.  So we only reload when a field applicable to the filters changes (inventory, zones, start, end).

src/app/campaign/store/models/flight-overlap.models.ts
```diff
@@ -0,0 +1,22 @@
...
+  error?: any;
+}
+
+export const overlapFilters = (flight: Flight): string => {
+  if (flight && flight.set_inventory_uri && flight.zones && flight.zones.length && flight.endAt && flight.startAt) {
+    return [
+      `inventory=${flight.set_inventory_uri.split('/').pop()}`,
+      `zones=${flight.zones.map(z => z.id).join(',')}`,
+      `before=${flight.endAt.toISOString()}`,
+      `after=${flight.startAt.toISOString()}`
```
cavis, line 18: I _think_ these before/after filters do what i want?  Since these are timestamps.

src/app/campaign/store/models/campaign-state.factory.ts
```diff
@@ -181,6 +181,13 @@ export const createFlightDaysState = () => ({
   }
 });
 
+export const createFlightOverlapState = () => ({
+  flightOverlap: {
+    ids: [],
+    entities: {}
+  }
+});
+
```
svevang, line 190: Overall nice succinct example of ngrx state/reducers/effects etc in this PR.

src/app/campaign/store/models/flight-overlap.models.ts
```diff
@@ -0,0 +1,22 @@
...
+  loading?: boolean;
+  error?: any;
+}
+
+export const overlapFilters = (flight: Flight): string => {
+  if (flight && flight.set_inventory_uri && flight.zones && flight.zones.length && flight.endAt && flight.startAt) {
+    return [
+      `inventory=${flight.set_inventory_uri.split('/').pop()}`,
+      `zones=${flight.zones.map(z => z.id).join(',')}`,
+      `before=${flight.endAt.toISOString()}`,
```
svevang: Seeing that the `before` filter (apps/model/flight.rb) is using the inclusive comparator `<=` in the `Flight#before` scope.  I _think_ we decided that we're using inclusive an end-at in the UI (e.g. calendar widget), which is the `endAtFudged`. If that's right than this should be `flight.endAtFudged`.

svevang (changes requested):
Just one change on end date. looks good!

src/app/campaign/store/models/flight-overlap.models.ts
cavis: Good catch!  Will update that.

cavis: Okay, after digging in a bit more, ~I think this is already what we want~.  And very confusing.

So I select `10/19` and `10/30` in the UI.  As-is, this branch fires off a request with filters:

```
before=2020-10-31T00:00:00.000Z
after=2020-10-19T00:00:00.000Z
```

Which get run through the Flight.rb scopes: 

```ruby
scope :before, ->(time) {
  where('flights.start_at <= ?', time)
}
scope :after, ->(time) {
  where('flights.start_at >= ? OR flights.end_at >= ?', time, time)
}
```

And we end up with SQL:

```sql
WHERE start_at <= '2020-10-31T00:00:00Z'
AND (start_at >= '2020-10-19T00:00:00Z' OR end_at >= '2020-10-19T00:00:00Z')
```

So any returned flights have to start _sometime_ <= midnight on the 31st.  Which is pretty close to what we want, if not exact.  (Ideally, we'd _exclude_ flights starting right at midnight?)

Hrmm...

svevang (approved):
Looks good!

PR #295, opened 2020-10-29
Patch the set_advertiser_uri field in all cases

cavis:
The advertiser dropdown/typeahead is not populating for existing campaigns.

I'm not sure I completely understand what's going on with this.  That form `patchValue` call appears to blank-out any text fields you don't include the keys for, which is sort of the opposite of what I'd expect from a patch.  Maybe that changed in the angular upgrade?

In any case ... the fix for this is to pattern match on the field.  If it looks like a typeahead "advertiser name" ... just leave it blank.  If it looks like a URI, include it.  For some reason that autocomplete component doesn't clear itself when it receives a blank update?

There are some other preexisting bugs with the invalid/changed status on this field as well.  (When you add a new advertiser, the field still looks invalid until you change something else in the form).  But maybe we'll tackle another day.  Or maybe the advertiser CRUD pages and get rid of this typeahead.

### DIFF
--- a/src/app/campaign/form/campaign-form.component.ts
+++ b/src/app/campaign/form/campaign-form.component.ts
@@ -87,8 +87,12 @@ export class CampaignFormComponent implements OnInit {
 
   updateCampaignForm(campaign: Campaign) {
     const { name, type, repName, notes, set_account_uri, set_advertiser_uri } = campaign;
-    const findAdvertiserByURI =
-      set_advertiser_uri && this.advertisers && this.advertisers.find(adv => adv.set_advertiser_uri === set_advertiser_uri);
+
+    // set_advertiser_uri might be a typeahead advertiser name instead
+    // NOTE: this fails if someone types a url as the advertiser name
+    const isAdvertiserName = set_advertiser_uri && !set_advertiser_uri.match(/^http(s)?:.+\/api\/v1\//);
+    console.log('updateCampaignForm', isAdvertiserName, set_advertiser_uri);
+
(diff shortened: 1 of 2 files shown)


src/app/campaign/form/campaign-form.component.ts
```diff
@@ -87,8 +87,11 @@ export class CampaignFormComponent implements OnInit {
 
   updateCampaignForm(campaign: Campaign) {
     const { name, type, repName, notes, set_account_uri, set_advertiser_uri } = campaign;
-    const findAdvertiserByURI =
-      set_advertiser_uri && this.advertisers && this.advertisers.find(adv => adv.set_advertiser_uri === set_advertiser_uri);
+
+    // set_advertiser_uri might be a typeahead advertiser name instead
+    // NOTE: this fails if someone types a url as the advertiser name
+    const isAdvertiserName = set_advertiser_uri && !set_advertiser_uri.match(/^http(s)?:.+\/api\/v1\//);
```
svevang, line 93: Makes sense. I wonder if we can proxy this field in the future somehow?  Like a `advertiser_name` field that proxies to the `set_advertiser_uri` key. Looks good!

svevang (approved):
Looks good!

src/app/campaign/form/campaign-form.component.ts
cavis: +1.  I didn't dig too far into that autocomplete component ... but it would be more straightforward if it didn't reuse this field to store whatever partial string you've typed.

PR #306, opened 2021-02-03
Fixup creative filtering

### LINKED ISSUE #304: Allow actuals outside of contract dates
Contract start/end dates should operate more or less independent of the actuals.

### LINKED ISSUE #305: Creatives filter field not working
I see the creatives text filter firing off to Augury, and getting back trimmed down results.  But the paged list doesn't actually update.  So that's a bug!

Also - not real useful to display urls on that page.  Since most things uploaded to `a.prxu.org` will start with the same prefix.  You can't tell them apart!  Should probably both (A) default that field to the filename not the full URL, and (B) text align right?

cavis:
- [x] Fixes #305 - had to dig around a bit to avoid a large diff here.  The Creative store here is basically a global id=>object hash.  And the creatives-list filters it based on the current page param.  I amended that to also store the last `text` filter an object came in with, and filter the creatives-list by `text` as well.
- [x] Fixes #304 - remove contract date validations entirely, as Augury doesn't have them.  And it appears these are basically reporting-only fields.
- [x] Also fixed a minor bug where adding creatives to a flight didn't mark it as "changed".

### DIFF
--- a/src/app/campaign/store/selectors/creative.selectors.ts
+++ b/src/app/campaign/store/selectors/creative.selectors.ts
@@ -32,10 +32,17 @@ export const selectCreativeById = createSelector(
 
 export const selectShowCreativeListRoute = createSelector(selectRouterState, state => state && state.url.indexOf('creative/list') > -1);
 export const selectCreativeListCurrentPage = createSelector(selectCreativeParams, params => params && params.page);
-export const selectCreativeListPageOrderedByCreatedAt = createSelector(selectAllCreatives, selectCreativeListCurrentPage, (states, page) =>
-  states
-    // filter out any 'new' temp creatives on the state and filter to creatives on current page
-    .filter(state => state && state.creative && state.creative.createdAt && ((!page && state.page === 1) || state.page === page))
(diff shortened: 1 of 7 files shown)


src/app/campaign/store/selectors/creative.selectors.ts
```diff
@@ -32,10 +32,17 @@ export const selectCreativeById = createSelector(
...
+export const selectCreativeListPageOrderedByCreatedAt = createSelector(
+  selectAllCreatives,
+  selectCreativeListCurrentPage,
+  selectCreativeListCurrentText,
+  (states, page, text) =>
+    states
+      // filter out any 'new' temp creatives on the state and filter to creatives on current page/text
+      .filter(state => state && state.creative && state.creative.createdAt && ((!page && state.page === 1) || state.page === page))
+      // bit hacky, since the store is overwritten on every remote call ... but apply current text filter
+      .filter(state => (!text && state.text === '') || state.text === text)
```
svevang, line 45: 👍

svevang (approved):
Looks good!

PR #307, opened 2021-02-03
Unique per booleans

cavis:
Route through the `uniquePerCampaign` and `uniquePerAdvertiser` booleans.  Those default to "true" in Augury, so they're actually inversed in this form (I called them `notUniquePerCampaign` and `notUniquePerAdvertiser` to be ridiculously specific about it).

image


svevang (approved):
Looks good! 👍

PR #327, opened 2021-03-25
#316: Remove FAB buttons from flight form

### LINKED ISSUE #316: Duplicate/Delete flight buttons shouldn't be FABs
As Rick pointed out, according to Material spec, we are incorrectly using FABs when we should use buttons instead for the Duplicate/Delete flight buttons.

rpeterman-gp:
Closes #316 

## This PR:
- Remove FAB buttons from flight form
- Adds context menu to campaign nav's flights' tabs to dup/delete active flight
- Fix border radius on new campaign button
- Update tests for flight form and campaign components

### DIFF
--- a/src/app/campaign/nav/campaign-nav.component.ts
+++ b/src/app/campaign/nav/campaign-nav.component.ts
@@ -31,9 +31,13 @@ import { Campaign, FlightState } from '../store/models';
         <mat-icon aria-hidden>file_copy</mat-icon>
         <span>Duplicate Flight</span>
       </button>
-      <button class="menuItem--warn" mat-menu-item (click)="onDelete()" *ngIf="canDeleteActiveFlight">
-        <mat-icon aria-hidden>delete</mat-icon>
+      <button class="menuItem--warn" mat-menu-item (click)="onDelete()" *ngIf="canDeleteActiveFlight && !softDeleted">
+        <mat-icon color="warn" aria-hidden>delete</mat-icon>
         <span>Delete Flight</span>
+      </button>
+      <button class="menuItem--primary" mat-menu-item (click)="onDelete()" *ngIf="softDeleted">
+        <mat-icon color="primary" aria-hidden>restore</mat-icon>
+        <span>Restore Flight</span>
       </button>
     </mat-menu>
   `,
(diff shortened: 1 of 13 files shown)


brandonhundt:
I'm having local dev issues, but what I did see looks good to me. @cavis Can you take a quick look?

cavis:
FWIW - this is what it looks like now (and you only get "Delete" for saved and delete-able flights). 
image

src/app/campaign/nav/campaign-nav.component.ts
```diff
@@ -19,20 +20,56 @@ import { Campaign, FlightState } from '../store/models';
...
     </mat-nav-list>
     <a class="secondary-link" routerLink="" (click)="createFlight.emit()"><mat-icon>add</mat-icon> Add a Flight</a>
+    <mat-menu panelClass="menuPanel" #flightMenu="matMenu">
+      <button mat-menu-item (click)="onDuplicate()">
+        <mat-icon aria-hidden>file_copy</mat-icon>
+        <span>Duplicate Flight</span>
+      </button>
+      <button class="menuItem--warn" mat-menu-item (click)="onDelete()" *ngIf="canDeleteActiveFlight">
+        <mat-icon aria-hidden>delete</mat-icon>
+        <span>Delete Flight</span>
```
cavis, line 36: I think this should toggle to "Restore Flight" if you've soft deleted it (clicked this, but haven't saved yet).

In fact, clicking this twice _does_ restore the flight.

src/app/campaign/status/campaign-status.component.ts
```diff
@@ -22,11 +22,11 @@ import { FlightState } from '../store/models';
       <grove-campaign-report [campaignName]="campaignName" [flights]="flights" [reportData]="reportData"></grove-campaign-report>
       <button mat-menu-item [disabled]="!canDuplicate" (click)="onDuplicate()">
         <mat-icon aria-hidden>file_copy</mat-icon>
-        <span>Duplicate</span>
+        <span>Duplicate Campaign</span>
```
cavis, line 25: +10 i think this was a source of confusion.

cavis (changes requested):
The text (and maybe icon?) should change when "Restoring" the flight.  But otherwise I think this looks good!

src/app/campaign/nav/campaign-nav.component.ts
rpeterman-gp: Great idea! The fact this is a toggle wen right over my head. Change has been pushed.

src/app/campaign/status/campaign-status.component.ts
rpeterman-gp: Added it to match the wording of the Delete Flight button. Thought it was best to be clear about what is being deleted.

cavis (approved):
👍 looks/works good to me!

PR #328, opened 2021-03-26
Dont load flights twice

cavis:
There's some code in the campaign service to `GET /api/v1/campaigns/123` with zoomed flights/flight-days.  It then checks if all the flights were present in that response (there's a `hal_api-rails` thing that only returns the first 10).  If not, it makes a 2nd request for `GET /api/v1/campaigns/123/flights?per=<total>` with zoomed flight-days.

Was seeing that 2nd request fire even when unnecessary.  This resolves that.

And adds a z-index so the material SnackBar is above the `prx-header`.

### DIFF
--- a/src/app/core/campaign/campaign.service.ts
+++ b/src/app/core/campaign/campaign.service.ts
@@ -21,14 +21,16 @@ export class CampaignService {
 
   loadCampaignZoomFlights(id: number): Observable<{ campaignDoc: HalDoc; flightDocs: HalDoc[] }> {
     return this.augury.follow('prx:campaign', { id, zoom: 'prx:flights' }).pipe(
-      switchMap(campaignDoc => campaignDoc.follow('prx:flights').pipe(map(flightDocs => ({ campaignDoc, flightDocs })))),
-      switchMap(({ campaignDoc, flightDocs }: { campaignDoc: HalDoc; flightDocs: HalDoc }) => {
+      switchMap(campaignDoc => campaignDoc.followItems('prx:flights').pipe(map(flightDocs => ({ campaignDoc, flightDocs })))),
+      switchMap(({ campaignDoc, flightDocs }: { campaignDoc: HalDoc; flightDocs: HalDoc[] }) => {
         // if total is greater than count, request all flights
-        let params: any;
-        if (+flightDocs['total'] > +flightDocs['count']) {
-          params = { per: +flightDocs['total'] };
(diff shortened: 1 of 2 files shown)


src/app/core/campaign/campaign.service.ts
```diff
@@ -21,14 +21,16 @@ export class CampaignService {
 
   loadCampaignZoomFlights(id: number): Observable<{ campaignDoc: HalDoc; flightDocs: HalDoc[] }> {
     return this.augury.follow('prx:campaign', { id, zoom: 'prx:flights' }).pipe(
-      switchMap(campaignDoc => campaignDoc.follow('prx:flights').pipe(map(flightDocs => ({ campaignDoc, flightDocs })))),
-      switchMap(({ campaignDoc, flightDocs }: { campaignDoc: HalDoc; flightDocs: HalDoc }) => {
+      switchMap(campaignDoc => campaignDoc.followItems('prx:flights').pipe(map(flightDocs => ({ campaignDoc, flightDocs })))),
```
cavis, line 24: Bit into the weeds of `HalDoc` ... but to get back an array of flight docs, you have to `followItems` which is a couple of chained `follow('prx:flights').followItems('prx:items')` to get the nested array.

So now this is correctly getting the array - but how to get the total/count that was [a few levels above](http://cav.is/lIUt4aj.png)?

Those attributes actually get set onto each of these `HalDocs` returned.  So if we got > 0 back, we can just check the first for the `.total()`.

src/app/core/campaign/campaign.service.ts
```diff
@@ -21,14 +21,16 @@ export class CampaignService {
...
-        if (+flightDocs['total'] > +flightDocs['count']) {
-          params = { per: +flightDocs['total'] };
+        if (flightDocs.length > 0 && flightDocs[0].total() > flightDocs.length) {
+          return campaignDoc
+            .followItems('prx:flights', { per: flightDocs[0].total() })
+            .pipe(map(docs => ({ campaignDoc, flightDocs: docs })));
+        } else {
+          return of({ campaignDoc, flightDocs });
         }
-        return campaignDoc.followItems('prx:flights', params).pipe(map(docs => ({ campaignDoc, flightDocs: docs })));
```
cavis, line 31: Calling `followItems` with params will always make a remote call, without checking `_embedded` zooms.  Moved this up into the `if` condition, so we're not always calling it.

kookster (approved):
lgtm!

